PR #68, opened 2021-01-15
Extract common component replacing ActionListFilters and StatusboardFilters

bbliem:
The respective filter components in the `actions` and `dashboard` parts of the UI were almost identical. I extracted a common component to replace both.


terotik (approved):

terotik:
Looks good

PR #75, opened 2021-01-21
Feature/contained themes

terotik:
Restructure themes. Move all the theme related files and assets under theme-named directory in `static/`
Makes handling theme related assets easier and prepares for possible submoduling of themes.


terotik:
Any concerns about this @juyrjola?

PR #78, opened 2021-02-03
Wip/category pages

bbliem:


components/contentblocks/CategoryListBlock.js
```diff
@@ -23,38 +23,12 @@ const CategoryListBlock = (props) => {
...
-        progress: '60',
-      },
-      {
-        id: '5',
-        imageUrl: placemarkerImageUrl,
-        title: 'Toimenpidekokonaisuus 5',
-        progress: '60',
-      },
-    ],
+    categories: props.categories.map(({ id, name, imageUrl }) => ({
```
terotik, line 26: This processing is probably unnecessary. We can just change the `{ cat.title }` into `{ cat.name }` below. And we can drop the `progress` for now.

PR #146, opened 2023-08-25
Refactor indicators page filters

woodwoerk:
Replace custom filters on the indicators page with actions filters. This fixes a few bugs on the indicators page (crashing when no items are in the table, incorrect category order/ nesting etc) and supports Durham who had issues with the category filtering.

Indicator filtering is still semi-hardcoded but this is a small step towards customisable indicator filters.

#### TODO
- [ ] Show the correct indicator count and label
- [ ] Hide categories that aren't in use (requested by Durham)


.prettierrc
woodwoerk, line 1: @juyrjola I couldn't get the `.mjs` config file to work so added this temporarily. Prettier complained about not finding a loader for mjs files, any idea on why that might be? Except some TS config (which didn't seem to help), I couldn't spot any mjs config differences compared to paths.

common/__generated__/possible_types.json
woodwoerk, line 1: Running GraphQL codegen made many more changes than I expected, why is this? Had it just not been ran for a while?

.prettierrc
juyrjola: No problemo, we can go with that for now. We need to harmonize the common code (mostly stuff related to our custom Node Express server) with paths-ui anyway at some point, so we can revisit then.

common/__generated__/possible_types.json
juyrjola: I'm guessing it's the large dependency version update we did on the backend side. That also changed how the gql schema is generated.

tituomin:
Not maybe part of this pull request specification but still there was the idea to make the hardcoded "Themes" column label in the indicator table also dynamic / use the category name from the GQL API

tituomin (approved):

components/indicators/IndicatorsHero.tsx
```diff
@@ -1,28 +1,32 @@
...
 import { useRouter } from 'next/router';
-import {
-  Container, Row, Col, Nav, NavItem,
-} from 'reactstrap';
+import { Container, Row, Col, Nav, NavItem } from 'reactstrap';
 import styled from 'styled-components';
 import { readableColor } from 'polished';
 import RichText from 'components/common/RichText';
-import { withTranslation } from '../../common/i18n';
+import { useTranslation, withTranslation } from '../../common/i18n';
```
terotik, line 8: Could change this to absolute path. This is an ancient component from the times we didn't support those yet.

terotik (commented):
All good. The nitpick comment not a dealbreaker :)

PR #152, opened 2023-09-01
Update indicator table to make indicators accessible by name

woodwoerk:
The indicator link was previously hard to find and raised by at least Äänekoski. This change makes the chevron icon a button to control expansion and the indicator name a link (consistent with our other tables).

A nice improvement here would be to remember the hierarchy expansion state between pages so that the user could easily navigate back and forth between nested indicators. We could think about this along side the bigger indicators redesign.

https://github.com/kausaltech/kausal-watch-ui/assets/15343658/3b6c461f-8743-4069-9f31-432159d0999e

### DIFF
--- a/components/common/Icon.js
+++ b/components/common/Icon.js
@@ -3,75 +3,74 @@ import PropTypes from 'prop-types';
 import SVG from 'react-inlinesvg';
 import { useTheme } from 'common/theme';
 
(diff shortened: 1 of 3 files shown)


components/common/Icon.js
woodwoerk, line 1: @terotik In the end I felt the original chevron icon looked better :D the filled triangle was too blocky/ heavy when repeated on every row. Is it still worth including the new icon I added or should I revert these `Icon` changes?

terotik: Agreed. Looks good!

PR #153, opened 2023-09-05
Updated node and yarn versions in README

lilia1891:
Updated node and yarn versions


terotik:
Thanks for the update!

PR #154, opened 2023-09-05
feat: support block anchor links 

woodwoerk:
Add an autogenerated id to all `StreamField` blocks. For now they just follow the block index e.g. `#block-1` represents the first block in a page, in future we could improve the editor experience by allowing editors to add their own id.

### DIFF
--- a/common/blocks.types.ts
+++ b/common/blocks.types.ts
@@ -1,0 +1,3 @@
+export interface CommonContentBlockProps {
+    id?: string
+}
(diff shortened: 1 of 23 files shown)


terotik (commented):
Seems to work great. I'm only wondering that to me `#block-n` sounds a bit technical? Do you think something like `#section-n` would sound friendlier @woodwoerk ?

Although as block is what we call them in the admin, maybe `block-n`  is easier for admins to remember the format.

woodwoerk:
> Seems to work great. I'm only wondering that to me `#block-n` sounds a bit technical? Do you think something like `#section-n` would sound friendlier @woodwoerk ?
> 
> Although as block is what we call them in the admin, maybe `block-n` is easier for admins to remember the format.

@terotik Good point, `section` is definitely less techy. Yeah, I chose block for consistency with the admin lingo, but since we'd need to document this anyway perhaps it's not an issue to diverge 🤔  let's go with `section`!

PR #156, opened 2023-09-07
Support themeable sticky navigation

woodwoerk:
- Use sticky navigation if defined in the theme
- Switch sticky nav to be the primary nav bar on small screens
- Render a nav spacer to fix jumping content while transitioning to sticky

Probably easier to review by commit, the first includes prettier fixes.

### DIFF
--- a/common/environment.ts
+++ b/common/environment.ts
@@ -1,0 +1,1 @@
+export const isServer = typeof window === 'undefined'
--- a/components/common/GlobalNav.js
+++ b/components/common/GlobalNav.js
@@ -1,8 +1,14 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import {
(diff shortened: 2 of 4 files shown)


terotik (changes requested):
Super thanks @woodwoerk ! A couple of things:

- Mobile size doesn't seem to expand on place. I'm ok disabling stickiness for hamburger nav. But if it's an easy fix to get the expansion working let's do that.
- Could we add a quick css transition on navbar appearance?

woodwoerk:
@terotik I've added a nav transition in but not out, sliding out would be trickier because the element is immediately rendered in a different place upon `isNavFixed` changing. It'd be possible with a delayed `setIsNavFixed` but might be finicky, please test it out and let me know :)

terotik (approved):
Fantastic. Works so nice now - we probably should make this as default!

PR #158, opened 2023-09-12
Fixed typehead layout on indicator insight page

lilia1891:
Bootstrap typehead breaks the layout displaying a clear button with a double cross outside the field.

typehead

* added import for react typehead CSS;
* added styles to Typehead to remove the redundant second cross


woodwoerk (approved):
Good catch, looks good to me!

terotik (approved):

PR #165, opened 2023-09-21
Truncate long category names in breadcrumbs

woodwoerk:
Truncate bread crumb categories longer than 90 characters, show the full category name in a tooltip.

Screenshot 2023-09-21 at 16 48 07

### DIFF
--- a/components/actions/ActionHero.tsx
+++ b/components/actions/ActionHero.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Container, Row, Col } from 'reactstrap';
+import { Container, Row, Col, UncontrolledTooltip } from 'reactstrap';
 import styled from 'styled-components';
 import { useTheme } from 'common/theme';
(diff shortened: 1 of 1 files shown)


components/actions/ActionHero.tsx
```diff
@@ -150,65 +151,98 @@ const ActionNumber = styled.span`
...
+      name: getCategoryName(cat, showIdentifiers),
+      url: getCategoryUrl(cat, primaryCT),
+      parent: cat.parent
+        ? {
+            id: cat.parent.id,
+            name: getCategoryName(cat.parent, showIdentifiers),
+            url: getCategoryUrl(cat.parent, primaryCT),
+          }
+        : undefined,
+    }));
```
woodwoerk, line 237: @terotik This is what I mentioned in the daily, I'm a big fan of using `map` and small helper functions with a clear input and output for these kinds of trickier data mappings.

PR #169, opened 2023-09-25
Fix/umbrella page category filter

tituomin:
Fixes problem with Umbrella page action list page categories not working


woodwoerk (approved):
Nice! Looks good :)

PR #170, opened 2023-09-25
Allow hiding indicator list page insights with page settings

tituomin:


components/indicators/IndicatorList.tsx
```diff
@@ -210,9 +210,10 @@ const getFirstUsableCategoryType = (categoryTypes, indicators) =>
 interface Props {
   title: string;
   leadContent: IndicatorListPage['leadContent'];
+  displayInsights: boolean;
```
woodwoerk, line 213: According to the generated types this could be undefined. Should we make this optional and default to `false` for backwards compatibility?

tituomin: Hmm. That is a problem in the type generation since in the Django backend it cannot be null/undefined and in the GraphQL schema it's a Boolean. But still, I can make the change.

tituomin: Fixed in https://github.com/kausaltech/kausal-watch-ui/pull/170/commits/71122a2b54c67d745fbc2f8e40419f816a330572

woodwoerk (approved):

PR #172, opened 2023-09-27
Added responsive styles to embeds

lilia1891:
* Individual embeds: added responsive styles to iframes according to the received CSS classes;
* RichText embeds: added styles for the received responsive object class.

### DIFF
--- a/components/common/RichText.tsx
+++ b/components/common/RichText.tsx
@@ -52,7 +52,7 @@ type RichTextImageProps = {
   };
 };
 
-const RichTextStyles = styled.div`
+const StyledRichText = styled.div`
   .responsive-object {
     position: relative;
   }
@@ -255,7 +255,7 @@ export default function RichText(props: RichTextProps) {
 
   return (
     <div {...rest} className={`text-content ${className || ''}`}>
-      <RichTextStyles>{parsedContent}</RichTextStyles>
+      <StyledRichText>{parsedContent}</StyledRichText>
     </div>
   );
 }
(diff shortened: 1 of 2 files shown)


components/common/RichText.tsx
```diff
@@ -52,6 +52,22 @@ type RichTextImageProps = {
   };
 };
 
+const RichTextStyles = styled.div`
```
tituomin, line 55: Because these styled components are still components, I wouldn't name it as just --Styles, even thought styling is here the only reason for the component's existence. I would maybe call these StyledRichText or RichTextContainer or something like that. (I think we have used at least Styled* elsewhere.

tituomin:
One minor naming suggestion. Otherwise, looks good, but maybe @terotik should take a look at the actual CSS.

tituomin (changes requested):

components/common/RichText.tsx
lilia1891: Changed. Thank you for a suggestion.

terotik (approved):
Looks good to me thanks!
`aspect-ratio` css is fairly new (to me at least) so I'm slightly concerned about the browser support. [Caniuse](https://caniuse.com/mdn-css_properties_aspect-ratio) gives 89.98% global support.

It's mainly the slightly older versions and IE who lack support. We could consider [this fallback](https://codepen.io/una/pen/BazyaOM) if we get reports of embeds not working in some browsers.

Ok to relase from me

tituomin (approved):
Good! There was another case with the same naming thing, otherwise great

PR #178, opened 2023-10-02
Implement provider-specific sizing for embeds

tituomin:


components/common/StreamField.tsx
```diff
@@ -289,19 +294,46 @@ const ResponsiveStyles = styled.div`
     height: 100%;
   }
 
-  .responsive-object-small iframe {
-    aspect-ratio: 16 / 9;
-    max-width: 300px;
+  .responsive-object-small {
```
woodwoerk, line 297: I'm assuming the `responsive-object` class is applied in addition to the sizing classes i.e. `<iframe class="responsive-object responsive-object-small" ... />`, is that the case?

woodwoerk (approved):
Looks good! Is this the end of the resizable embeds saga? 😁

components/common/StreamField.tsx
tituomin: Yes!

tituomin: @woodwoerk I actually came up with an improvement myself. Would it be bette if I switched these selectors around:
`  .responsive-object-large {
    &.responsive-object[data-embed-provider='${EmbedProvider.YOUTUBE}']`
    
  There would be less repetition of the data-embed-providers. So I mean the enclosing selector would be the .responsive-object selector and they would be "grouped by" provider.

tituomin: I mean:
`.responsive-object[data-embed-provider='${EmbedProvider.YOUTUBE}'] { &.responsive-object-large { [...]`

woodwoerk: Nice! Yeah that's much clearer. Super nitpick: you could avoid the repetition of `aspect-ratio` by placing that outside of the `-<size>` specific classes, but this approach may be more readable anyway!

woodwoerk (approved):

PR #179, opened 2023-10-02
Display status in actiondashboard phase tooltip

terotik:
https://github.com/kausaltech/kausal-watch-ui/assets/664877/4d79ab6e-6a11-42a7-b66a-a83d7d6cf4c8

### DIFF
--- a/components/dashboard/ActionTableTooltips.js
+++ b/components/dashboard/ActionTableTooltips.js
@@ -108,16 +108,18 @@ export const phasesTooltipContent = (
   plan
 ) => {
   const getMergedName = (mergedWith, planId) => {
-    if (mergedWith.plan.id !== planId)
+    if (mergedWith.plan.id !== planId) {
       return `${mergedWith.plan.shortName} ${mergedWith.identifier}`;
-    else return mergedWith.identifier;
+    } else {
+      return mergedWith.identifier;
+    }
   };
 
   const statusDisplay = (
     <StatusLabel color={status?.color}>{status.label}</StatusLabel>
   );
   // If action is merged, display merged status
-  if (merged)
+  if (merged) {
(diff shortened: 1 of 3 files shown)


terotik:
Requesting @tituomin review as you have worked on phases/statuses and not sure if there are edge cases I have not handled here. I didn't have time to add any type checking sorry :/

components/dashboard/ActionTableTooltips.js
```diff
@@ -85,35 +113,32 @@ export const phasesTooltipContent = (
...
-    return (
-      <div>
-        <h5>{status.name}</h5>
-      </div>
-    );
-  return hasImplementationPhases ? (
+  // If action has no active phase or it's cancelled, or plan has no implementation phases : display only status
+  if (
+    !activePhase ||
+    status?.identifier === 'cancelled' ||
```
tituomin, line 132: Since this file isn't type checked yet, we can leave it be, but in the future we can check for legal status summary identifier by comparing to https://github.com/kausaltech/kausal-watch-ui/blob/feat%2Fshow-status-on-tooltip/common/__generated__/graphql.ts#L773-L785

components/dashboard/ActionTableTooltips.js
```diff
@@ -85,35 +113,32 @@ export const phasesTooltipContent = (
...
+    !hasImplementationPhases
+  )
+    return statusDisplay;
```
tituomin, line 135: would maybe use { } after if everywhere.

tituomin (approved):

components/dashboard/ActionTableTooltips.js
terotik: Good point. Done!

PR #180, opened 2023-10-02
Improve area graphs and use trend line visibility setting

terotik:
Improve drawing indicator graphs if `graphs.areaGraphs=true` in theme (should be in backend)
Also implement `graphs.showTrendline=false` theme setting (should be in backend)
From this:
Screenshot 2023-10-02 at 22 05 22
to this (do not fill if multiple traces)
Screenshot 2023-10-02 at 22 05 04
http://klima-winterthur.localhost:3000/indicators/1687

And from this:
Screenshot 2023-10-02 at 22 06 55
to this (no trend line, fill also goal as per customer request):
Screenshot 2023-10-02 at 22 07 18
http://klima-winterthur.localhost:3000/indicators/2056

### DIFF
--- a/components/graphs/IndicatorGraph.tsx
+++ b/components/graphs/IndicatorGraph.tsx
@@ -4,6 +4,7 @@ import dynamic from 'next/dynamic';
 import { merge, uniq } from 'lodash';
 import styled from 'styled-components';
 import { useTheme } from 'common/theme';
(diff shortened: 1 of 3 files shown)


woodwoerk (approved):
Nice 🔥 I haven't tested locally but the code and examples look good

PR #181, opened 2023-10-05
Add alternative layout for frontPageHeroBlock

terotik:
Use small_image layout option for alternative version of FrontPageHeroBlock

From full width ui force cropped hero image
klima-winterthur watch-test kausal tech_

To container width hero image keeping the original image proportions
klima-winterthur localhost_3000_


tituomin:
I'll let @woodwoerk review unless he cannot and it's super urgent.

components/contentblocks/FrontPageHeroBlock.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import HeroFullImage from 'components/home/HeroFullImage';
+import HeroSmallImage from 'components/home/HeroSmallImage';
+
+interface FrontPageHeroProps {
+  id: string;
+  layout: 'small_image' | 'big_image';
```
woodwoerk, line 7: Nitpick: Would `full_width` and `contained` be clearer? An image that could have a height up to 1600px also sounds like a big image :D

components/home/HeroFullImage.tsx
```diff
@@ -0,0 +1,196 @@
+import { useTheme } from 'common/theme';
```
woodwoerk, line 1: Is there an old `HeroFullImage.js` which should be deleted?

woodwoerk (approved):
Nice work, looks good to me! Dropped one nitpick and one question 🕺 

By the way, I tend to mark very small things as nitpicks which are up to the author to decide whether they make sense or not – usually just coding conventions or naming of stuff. Feel free to ignore them if you disagree or if the time investment wouldn't be worth the change.

components/contentblocks/FrontPageHeroBlock.tsx
terotik: Absolutely would. Only reason I ended up using this is that we happened to have a legacy Small Image layout option in admin that had no UI effect anymore. I repurposed the unused backend choice for this. 

Small image kiiind of makes sense as in there the image is limited to container width instead of expanding to full browser width :D The height limit for rendition is arbitary as I would hope that no customer would try to put portrait ratio image there...

components/home/HeroFullImage.tsx
terotik: Good catch - I thought I used rename but it's still lingering there

PR #182, opened 2023-10-05
Fix site metadata

juyrjola:
- Set missing metadata tags in Layout
- Add missing baseURL to SiteContext


terotik (approved):

terotik:
Looks good to me. I didn't dive into meta/og formatting in detail, but trust that it works :)

PR #184, opened 2023-10-06
Support customised action column tables

woodwoerk:
I ended up refactoring a fair bit of this code but some components have just been separated/ moved:

- Table cell components have been separated under `components/dashboard/cells`
- Column configuration has been standardised under `COLUMN_CONFIG`, which defines functions to render cells, headers, tooltips and control sorting based on the column type
- Rendering of tooltips has been moved to the `ActionTableRow` level and tooltips now uncontrolled. Previously tooltips were controlled and rendered in `ActionStatusTable`, causing a rerender of the whole table whenever tooltips were opened/ closed
- Broken accessible tooltip markup has been removed, we were using the `hidden` HTML attribute which hides elements from all presentations including screen readers. Accessible tooltips to be added later.

Testable in staging: 

- [Häme](https://hame-ilmasto.watch.staging.kausal.tech/actions?view=dashboard) with organisation and custom name column label
- [Tampere](https://tampere-ilmasto.watch.staging.kausal.tech/actions?view=dashboard)
- [Bremen](https://bremen-klima.watch.staging.kausal.tech/actions?view=dashboard) stress test 

---
  - https://app.asana.com/0/0/1205007804913903


components/dashboard/ActionTableTooltips.tsx
```diff
@@ -23,13 +31,13 @@ const PhasesTooltipList = styled.ul`
   list-style: none;
 `;
 
-const PhasesTooltipListItem = styled.li`
-  color: ${(props) => (props.active === 'true' ? '#333' : '#ccc')};
+const PhasesTooltipListItem = styled.li<{ active: boolean }>`
```
juyrjola, line 34: I believe custom props (i.e. not passed directly to the DOM element) for styled components should be prefixed with a `$`, otherwise there will be complaints in the browser console: https://styled-components.com/docs/basics#passed-props

components/dashboard/ActionTableTooltips.tsx
```diff
@@ -23,13 +31,13 @@ const PhasesTooltipList = styled.ul`
...
+  color: ${(props) => (props.active ? '#333' : '#ccc')};
 `;
 
 const TaskTooltip = styled.div``;
 
-const StatusLabel = styled.div`
+const StatusLabel = styled.div<{ color: string }>`
```
juyrjola, line 40: Maybe the same thing here – although `color` does exist as a DOM prop, too

components/dashboard/cells/ImplementationPhaseCell.tsx
```diff
@@ -0,0 +1,30 @@
+import { ActionListAction } from '../dashboard.types';
+import styled from 'styled-components';
+import { PlanContextFragment } from 'common/__generated__/graphql';
+import ActionPhase from 'components/actions/ActionPhase';
+
+interface Props {
```
juyrjola, line 6: Would it be better to use `type` instead of `interface`? I bumped into this styleguide over the weekend, and they recommend `type OwnProps`: https://docs.twenty.com/developer/frontend/style-guide#props

They seemed to have pretty reasonable other practices, too. We might want to consider adopting some of them.

components/dashboard/cells/ImplementationPhaseCell.tsx
```diff
@@ -0,0 +1,30 @@
...
+const StatusDisplay = styled.div`
+  padding: ${(props) => props.theme.spaces.s050};
+  height: 100%;
+`;
+
+const ImplementationPhaseCell = ({ action, plan }: Props) => (
+  <StatusDisplay>
+    <ActionPhase
+      action={action}
+      status={action.statusSummary}
```
juyrjola, line 20: Does it makes sense to pass both `action` and then properties from inside `action` separately?

components/dashboard/cells/IndicatorsCell.tsx
```diff
@@ -0,0 +1,43 @@
+import { ActionListAction } from '../dashboard.types';
+import styled, { useTheme } from 'styled-components';
+import Icon from 'components/common/Icon';
+
+interface Props {
```
juyrjola, line 5: Ditto here.

components/dashboard/cells/OrganizationCell.tsx
```diff
@@ -0,0 +1,31 @@
...
+const OrganizationCell = ({ action }: Props) => {
+  if (!action.primaryOrg) {
+    return null;
+  }
+
+  return (
+    <OrgLogo
+      src={
+        action.primaryOrg.logo?.rendition?.src ??
+        '/static/themes/default/images/default-avatar-org.png'
```
juyrjola, line 23: This might need to have the `basePath` prefix to work properly when we are deploying under a path prefix.

juyrjola (commented):
Finishing the review early so the comments get visible

components/orgs/OrgContent.tsx
```diff
@@ -167,6 +168,11 @@ const GET_ORG_DETAILS = gql`
     plan(id: $planId) {
       ...OrgContentPlan
     }
+    # Org actions table uses the custom column config from the actions list page
+    planPage(plan: $planId, path: "/actions") {
```
juyrjola, line 172: Maybe use `plan.actionListPage` instead?

juyrjola:
Other than the issues I commented on, LGTM

components/dashboard/cells/OrganizationCell.tsx
woodwoerk: This is old code, but it could be the case that the primaryOrg logo has never been null so this hasn't been discovered

components/dashboard/cells/ImplementationPhaseCell.tsx
woodwoerk: I like this! We're following many of those already, and [others just make sense](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces). I did some reading up on `interface` vs `type` and felt this is the main benefit of types over interfaces:

- Interfaces with the same name in the same scope [merge declarations](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces), whereas duplicate types will throw an error. While rare, there's a possibility for bugs/ unpredictability if we accidentally declare duplicate interfaces.

PR #193, opened 2023-10-13
Quickfix/action filters alignment

terotik:
Before
Screenshot 2023-10-13 at 14 08 12

After
Screenshot 2023-10-13 at 14 07 11

Just in: contrast of dropdown active items (customer request)
Before
Screenshot 2023-10-13 at 14 33 40

After
Screenshot 2023-10-13 at 14 32 54

### DIFF
--- a/components/actions/ActionListFilters.tsx
+++ b/components/actions/ActionListFilters.tsx
@@ -168,6 +168,12 @@ const MainCategory = styled.div`
   }
 `;
 
+const FilterColumn = styled(Col)`
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+`;
+
 function sortDepthFirst<Type>(
   items: Type[],
   compareItem: (a: Type, b: Type) => number,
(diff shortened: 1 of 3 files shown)


components/actions/ActionListFilters.tsx
```diff
@@ -168,6 +168,12 @@ const MainCategory = styled.div`
   }
 `;
 
+const FilterColumn = styled(Col)`
```
woodwoerk, line 171: Juha shared a nice [styleguide](https://docs.twenty.com/developer/frontend/style-guide#styling) that recommends using the `Styled` prefix i.e. `StyledFilterColumn` to easily differentiate between styled components and _actual_ components. I kinda dig it, should we start following that convention?

components/common/SelectDropdown.tsx
```diff
@@ -81,11 +81,15 @@ function getSelectStyles<Option extends SelectDropdownOption>(
       const { indent } = data;
       const ret = {
         ...provided,
-        color: theme.themeColors.black,
+        color: isSelected
+          ? theme.themeColors.white
+          : isFocused
+          ? theme.themeColors.black
+          : theme.themeColors.black,
```
woodwoerk, line 88: Is the second ternary operator needed? Or should those colours be different depending on focus?

```suggestion
        color: isSelected
          ? theme.themeColors.white
          : theme.themeColors.black
```

woodwoerk (approved):
Nice improvements! Dropped two nitpicks

PR #200, opened 2023-10-25
Prevent plotly from clipping markers on axes

woodwoerk:
Also updated the Husky pre-commit config to do properly load Node when committing via the VSCode UI

Screenshot 2023-10-25 at 14 05 07


terotik (approved):
Good find that setting!

PR #203, opened 2023-10-26
fix: navigate to active plan home rather than root umbrella plan

woodwoerk:
Rather than navigating to the root umbrella plan, navigate to the current plan home under the secondary navigation "Home" link


terotik (approved):
looks good 😊

PR #206, opened 2023-10-30
Adapt to changes to IndicatorGroupBlock in backend

bbliem:
This is the frontend part of the changes in https://github.com/kausaltech/kausal-watch-private/pull/76. Unfortunately the schema changes due to the backend changes, so we cannot deploy the backend without breaking the frontend and vice versa. Perhaps we should add a shim, but that's always a PITA.

### DIFF
--- a/common/__generated__/graphql.ts
+++ b/common/__generated__/graphql.ts
@@ -444,7 +444,6 @@ export type ActionIndicator = {
   indicator: Indicator;
 };
 
-/** An enumeration. */
 export enum ActionIndicatorEffectType {
   /** decreases */
   Decreases = 'DECREASES',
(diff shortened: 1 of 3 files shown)


woodwoerk:
@bbliem Sorry I didn't make this clearer, but the frontend is already completed. PR here along with the other request and some type conversions: https://github.com/kausaltech/kausal-watch-ui/pull/197.

> Unfortunately the schema changes due to the backend changes, so we cannot deploy the backend without breaking the frontend and vice versa.

Could we schedule the frontend and backend to be deployed during off-peak hours?

bbliem:
Sorry, you probably mentioned this and I forgot. Anyway it has hardly any work, so I'm going to close this PR. I added a shim so we can upgrade the backend without breaking the UI. No need for a coordinated deployment now. (But the backend needs to be deployed before the UI.)

PR #207, opened 2023-10-31
Accommodate multiple logos on custom bgcolor in footer

terotik:
** requires theme update **
https://github.com/kausaltech/kausal-themes/pull/7

- Use new theme variables to enable different footer colors for secondary footer content (Statement, Funding logos, Other logos)
- Style more than 3 extra logos smaller

Screenshot 2023-10-31 at 18 16 13

Side effects:
If second footer contains statement AND otherLogos they are not adjacent anymore:
 
Screenshot 2023-10-31 at 18 25 02

Needs more work to handle this edge case, but should be ok for now

---
  - https://app.asana.com/0/0/1205750873498146


woodwoerk:
Sorry for the late review, but looks good ✅ I agree that the side effect is fine for now

PR #211, opened 2023-11-06
Feature/rationalize status colors

tituomin:
There was a problem that the action dashboard showed different colors from the normal action dashboard list because the latter didn't receive the full action object. This has been consistently fixed everywhere.

### DIFF
--- a/common/ActionStatusSummary.ts
+++ b/common/ActionStatusSummary.ts
@@ -49,24 +49,25 @@ export interface ActionWithStatusSummary {
   color?: string | null;
 }
 
+const DEFAULT_COLOR = 'grey050';
+
 export const getStatusColorForAction = (
   action: ActionWithStatusSummary,
   plan: PlanContextType,
   theme: Theme
 ) => {
-  const { color } = action;
+  const { color, statusSummary } = action;
   if (color != null) {
     return getThemeColor(color, theme);
   }
-  const { statusSummary } = action;
-  if (statusSummary == null) {
-    throw new Error('Action data is missing statusSummary');
+  if (
+    statusSummary == null ||
+    (statusSummary.color == null && statusSummary.identifier == null)
+  ) {
+    return getThemeColor(DEFAULT_COLOR, theme);
   }
   if (statusSummary.color != null) {
     return getThemeColor(statusSummary.color, theme);
-  }
-  if (statusSummary.identifier == null) {
-    throw new Error('Action data is missing statusSummary identifier');
   }
   const statusSummaryWithColor = getStatusSummary(plan, statusSummary);
   return getThemeColor(statusSummaryWithColor.color, theme);
--- a/common/ActionStatusSummary.ts
+++ b/common/ActionStatusSummary.ts
@@ -36,16 +36,17 @@ const getCacheKey = (
 
 export const getStatusSummary = memoize(_getStatusSummary, getCacheKey);
 
-export const getStatusColor = (color: string, theme: Theme) => {
+export const getThemeColor = (color: string, theme: Theme) => {
   return theme.graphColors[color];
 };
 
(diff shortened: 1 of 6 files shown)


common/ActionStatusSummary.ts
```diff
@@ -36,27 +36,38 @@ const getCacheKey = (
...
 
 export const getStatusColorForAction = (
   action: ActionWithStatusSummary,
   plan: PlanContextType,
   theme: Theme
 ) => {
-  const c = action?.color;
-  if (c != null) {
-    return getStatusColor(c, theme);
+  const { color } = action;
```
woodwoerk, line 57: Tiny nitpick: I think it's a bit clearer if we only destructure the action object once

```suggestion
  const { color, statusSummary } = action;
```

common/ActionStatusSummary.ts
```diff
@@ -36,27 +36,38 @@ const getCacheKey = (
...
+  if (color != null) {
+    return getThemeColor(color, theme);
   }
-  const s = action?.statusSummary;
-  if (s == null || (s?.identifier == null && s?.color == null)) {
+  const { statusSummary } = action;
+  if (statusSummary == null) {
```
woodwoerk, line 62: No need to change this since it was pre-existing, but in what scenario would `statusSummary` be null/ undefined? Throwing errors without rigid try / catch blocks in place gives me the heebie-jeebies. Here we could fall back to a default colour and log a Sentry error instead.

common/ActionStatusSummary.ts
```diff
@@ -36,27 +36,38 @@ const getCacheKey = (
...
     throw new Error('Action data is missing statusSummary');
   }
-  const statusSummary = s?.color == null ? getStatusSummary(plan, s) : s;
-  return getStatusColor(statusSummary.color, theme);
+  if (statusSummary.color != null) {
+    return getThemeColor(statusSummary.color, theme);
+  }
+  if (statusSummary.identifier == null) {
+    throw new Error('Action data is missing statusSummary identifier');
+  }
```
woodwoerk, line 70: Is this necessary?

components/actions/ActionsTable.js
```diff
@@ -54,13 +54,9 @@ const ACTION_ROW_FRAGMENT = gql`
...
   const checkedStatus = cleanActionStatus(action, plan.actionStatuses);
-
-  return (
-    <StatusBadge
-      plan={plan}
-      statusSummary={getStatusSummary(plan, action.statusSummary)}
-    />
-  );
+  const statusSummary = getStatusSummary(plan, action.statusSummary);
+  const actionWithStatusSummary = Object.assign({}, action, { statusSummary });
```
woodwoerk, line 58: Nitpick: They're very similar in practice, but I tend to prefer spreading objects over `Object.assign`. Mainly because it's easy to accidentally mutate an object by forgetting the empty object as the first argument, whereas object spreading always creates a new one. I think it's a bit more readable too, but that's personal preference

```suggestion
  const actionWithStatusSummary = { ...action, statusSummary };
```

woodwoerk (approved):
Nice work, `getStatusColorForAction` is much more readable now. I dropped a few nitpicks but feel free to merge this anyway.

common/ActionStatusSummary.ts
tituomin: Yeah, it would be a programming error, some obscure graphql query where we would have forgotten to add the field.You're right, Sentry was the only point to have this, but we haven't received any errors. I'll just skip the throwing altogether.

components/actions/ActionsTable.js
tituomin: We have been using Object.assign elsewhere, I'll leave this now but let's reach a conclusion later.

PR #224, opened 2023-11-24
Get indicator showcase bar colors from theme

terotik:
Also fix misaligned label on very short goal bars

before:
Screenshot 2023-11-27 at 16 53 29
afters:
Screenshot 2023-11-27 at 17 03 41

---
  - https://app.asana.com/0/0/1206010535152027

### DIFF
--- a/components/indicators/IndicatorProgressBar.js
+++ b/components/indicators/IndicatorProgressBar.js
@@ -2,6 +2,7 @@
 import React, { useContext, useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import styled, { ThemeContext } from 'styled-components';
+import { readableColor } from 'polished';
 import { motion, useAnimation, animate } from 'framer-motion';
 import { useTranslation } from 'common/i18n';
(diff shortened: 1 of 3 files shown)


components/indicators/IndicatorProgressBar.js
```diff
@@ -433,6 +434,7 @@ function IndicatorProgressBar(props) {
                   value={formatValue(startValue, i18n.language, minPrecision)}
                   unit={unit}
                   locale={i18n.language}
+                  negative={readableColor(startColor) === '#ffffff'}
```
woodwoerk, line 437: This feels like it could be prone to error (for example if `readableColor` is updated and starts returning `'#FFF'` instead), could we pass a `color` prop to ValueGroup instead? No need to do this now though

woodwoerk (approved):

components/indicators/IndicatorProgressBar.js
woodwoerk: Or we could define both the background colour and text colour in the theme

PR #225, opened 2023-11-29
Move timeliness labels to frontend and improve human readableness

woodwoerk:
image
image
image


tituomin:
I can't find the accept button on mobile but I accept! 🙂

PR #227, opened 2023-11-29
Add new phase timeline visualisation

woodwoerk:
Aligns all timeline visualisations to use a new `PhaseTimeline` component, which supports three layouts: 
- Horizontal
- Vertical (for small screen sizes and tooltips)
- Mini (for tables, with only the active phase label displayed)

stpaul-carp localhost_3000_actions_13
surrey-ccas localhost_3000_actions_view=dashboard
|  tampere-ilmasto localhost_3000_actions_1 1 7 (1) |  tampere-ilmasto localhost_3000_actions_1 1 7 |  klima-winterthur localhost_3000_actions_E1 1 |
|---|---|---|


terotik:
Looks good! A couple of questions:
- Status is now not visible on dashboard table by default. Should we add the status column in all plans, or augment phase cell with status information somehow if there is no separate column? It could be confusing if we show status pie chart but not statuses on the table.
- Contrast: I think `cardBackground.secondary` could be`grey005` by default. it feels the bg on this one should be as light as possible. Also getting contrast a11y error on Wave on the disabled/future phase labels.
- Not keen on status label changing location on Action View depending on if there is explanation. But I can not think a better solution. As the label without explanation is quite nicely placed where it's now.

terotik (approved):

PR #229, opened 2023-12-05
Added Other header for uncategorized actions and translation

lilia1891:


components/actions/ActionCardList.js
```diff
@@ -102,14 +103,20 @@ function ActionCardList(props) {
...
+    theme,
+    t
+  );
 
   return (
     <ActionsList>
       {groups.map((group) => (
         <ActionGroup key={group.id} tag="li">
-          {groups.length > 1 && (
+          {(groups.length > 1 || group.id === 'zzzz') && (
```
woodwoerk, line 119: We could store this as a variable e.g. `OTHER_GROUP_ID` and reference it here and on lines 90 & 95. That should make the code a bit more readable and future devs won't just think this group is sleeping 💤

locales/en-AU/common.json
```diff
@@ -5,5 +5,6 @@
   "responsible-parties": "Responsible organisations",
   "responsible-parties-missing": "No named responsible organisations",
   "primary-organization": "Organisation",
-  "organizations": "Organisations"
+  "organizations": "Organisations",
+  "other": "Other"
```
woodwoerk, line 9: We don't need to add translations to country specific variants of the same language (`en-AU`, `en-GB` etc) if the translations are the same. If it doesn't exist here it'll just fall back to `en` 🇬🇧

woodwoerk (approved):
Nice! Added a few nitpicks as comments

PR #243, opened 2023-12-21
Add AnnouncementBanner component to display sitewide announcements

tituomin:
Screenshot from 2023-12-21 18-19-07

### DIFF
--- a/components/common/AnnouncementBanner.tsx
+++ b/components/common/AnnouncementBanner.tsx
@@ -1,13 +1,13 @@
 import styled from 'styled-components';
 import { Container } from 'reactstrap';
 
-export const AnnouncementBannerSection = styled.div`
+export const StyledAnnouncementBannerSection = styled.div`
   padding: ${(props) => props.theme.spaces.s100} 0;
   background-color: ${(props) => props.theme.graphColors.blue070};
   color: ${(props) => props.theme.graphColors.grey010};
 `;
 
-export const AnnouncementBannerWrapper = styled.div`
+export const StyledAnnouncementBannerWrapper = styled.div`
   display: flex;
   justify-content: space-between;
   padding: 0;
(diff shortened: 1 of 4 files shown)


components/common/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,49 @@
+import styled from 'styled-components';
+import { Container } from 'reactstrap';
+
+export const AnnouncementBannerSection = styled.div`
```
woodwoerk, line 4: Nitpick: Juha and I agreed to started prefixing styled component names with `Styled`, i.e. `StyledAnnouncementBannerSection`. It makes names a bit verbose but is really handy to easily spot "real" components and styled components in the render function

components/common/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }
+  p ~ p {
+    margin-top: ${(props) => props.theme.spaces.s050};
+  }
+  a {
+    color: ${(props) => props.theme.themeColors.white};
+    font-weight: bold;
+  }
+  @media (max-width: ${(props) => props.theme.breakpointMd}) {
+  }
```
woodwoerk, line 28: Unnecessary breakpoint?

woodwoerk (approved):
Nice! Looking good, added a few nitpicks but nothing that prevents merging – we can decide upon the themed background colour later as more customers take this into use

components/common/AnnouncementBanner.tsx
tituomin: https://github.com/kausaltech/kausal-watch-ui/pull/243/commits/343691cd199f6f6e02ac8f28526d0760fb084fc5

components/common/AnnouncementBanner.tsx
tituomin: https://github.com/kausaltech/kausal-watch-ui/pull/243/commits/343691cd199f6f6e02ac8f28526d0760fb084fc5

PR #248, opened 2024-01-17
Allow no goal value on (normalized) indicator showcase

terotik:
Screenshot 2024-01-17 at 17 20 16

Handle edge case where goal value is not available:

Screenshot 2024-01-17 at 17 20 26

---
  - https://app.asana.com/0/0/1206202710087677

### DIFF
--- a/components/common/Switch.js
+++ b/components/common/Switch.js
@@ -7,7 +7,13 @@ function Switch(props) {
   return (
     <Form>
       <FormGroup switch>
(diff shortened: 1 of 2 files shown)


terotik:
Added outline on switch at the same time. Visibility of this has been mentioned before.

Screenshot 2024-01-17 at 17 32 35

lilia1891 (approved):

PR #251, opened 2024-01-22
Fix several problems with dashboard action status donuts

tituomin:
Screenshot from 2024-01-22 17-28-02
Screenshot from 2024-01-22 17-27-55
Screenshot from 2024-01-22 17-27-44
Screenshot from 2024-01-22 17-27-33

0. Refactor the phase donut implementation.

1. Don't count "inactive" actions as "no phase set"

2. Try to display in a clearer way that the percentage in the status donut does take into account the actions without any status set (not counted towards 100%, a question mark is displayed for that sector.) So the 73% in the example means that of all actions with a status set, 73% are not late or otherwise. Still, of all the actions, 1/4 doesn't have a status set at all.

### DIFF
--- a/common/preprocess.ts
+++ b/common/preprocess.ts
@@ -62,6 +62,22 @@ const cleanActionStatus = (action, actionStatuses) => {
   return newStatus;
 };
 
+class DonutSector {
+  count: number;
+
+  constructor(
+    public label: string,
+    public color: string,
+    public includeInTotal: boolean
+  ) {
+    this.count = 0;
+  }
+  increment() {
+    this.count++;
+    return this;
+  }
+}
+
 /*
  Process a list of actions and return an ordered list of statuses for statistics
  */
@@ -101,8 +117,8 @@ const getStatusData = (
           : label || unknownLabelText
       );
       progress.colors.push(theme.graphColors[colors.get(identifier) ?? color]);
-      if (sentiment == Sentiment.Positive) {
-        progress.good = progress.good + statusCount;
+      if (sentiment == Sentiment.Positive || identifier == 'NOT_STARTED') {
+        progress.good += statusCount;
       }
     }
     if (identifier !== ActionStatusSummaryIdentifier.Undefined) {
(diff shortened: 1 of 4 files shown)


tituomin:
Please note this implementation lacks translations. I will add them before merging.

terotik:
I know it was originally my suggestion, but wondering now if the ? is still a bit confusing. Maybe replace with '-' or leave empty? Question mark somehow seems to imply that we don't know the amount of no-status actions.

Might need some design exploration in the future still, but I think this is what we are able to achieve with donut chart for now.

tituomin:
> I know it was originally my suggestion, but wondering now if the ? is still a bit confusing. Maybe replace with '-' or leave empty? Question mark somehow seems to imply that we don't know the amount of no-status actions.
> 
> Might need some design exploration in the future still, but I think this is what we are able to achieve with donut chart for now.

Maybe you're right, just leaving it empty might actually be pretty clean.

common/preprocess.ts
```diff
@@ -62,6 +62,31 @@ const cleanActionStatus = (action, actionStatuses) => {
...
+    this.label = label;
+    this.includeInTotal = includeInTotal;
+    this.count = 0;
+  }
+  incr() {
+    this.count++;
+    return this;
+  }
+}
+
```
woodwoerk, line 89: No need to change this but just FYI if you weren't aware, JS classes allow us to define class properties within the constructor so there's no need to define the type on the class and also set the value separately:

```ts
class DonutSector {
  count: number;

  constructor(
    public label: string,
    public color: string,
    public includeInTotal: boolean
  ) {
    this.count = 0;
  }

  incr() {
    this.count++;
    return this;
  }
}

new DonutSector('foo', 'green', false)
```

common/preprocess.ts
```diff
@@ -62,6 +62,31 @@ const cleanActionStatus = (action, actionStatuses) => {
...
+    label: string;
+    color: string;
+    includeInTotal: boolean;
+  }) {
+    this.color = color;
+    this.label = label;
+    this.includeInTotal = includeInTotal;
+    this.count = 0;
+  }
+  incr() {
```
woodwoerk, line 84: Nitpick: I prefer clarity over saving characters in frontend code i.e. `increment()`

common/preprocess.ts
```diff
@@ -101,15 +128,35 @@ const getStatusData = (
           : label || unknownLabelText
       );
       progress.colors.push(theme.graphColors[colors.get(identifier) ?? color]);
-      if (sentiment == Sentiment.Positive) {
-        progress.good = progress.good + statusCount;
+      if (sentiment == Sentiment.Positive || identifier == 'NOT_STARTED') {
```
woodwoerk, line 131: Nitpick: It's a good practice to use strict equality in JS (`===`). The only exception being checking for `null` or `undefined` where `something != null` comes in handy.

common/preprocess.ts
```diff
@@ -101,15 +128,35 @@ const getStatusData = (
...
     }
     if (identifier !== ActionStatusSummaryIdentifier.Undefined) {
       totalCount += statusCount;
     }
   });
   progress.total = `${Math.round((progress.good / totalCount) * 100)}%`;
+  for (let i = 0; i < progress.values.length; i++) {
+    const label = progress.labels[i];
+    let value: number | string;
+    value = progress.values[i];
```
woodwoerk, line 143: Is value changed?

```suggestion
    const value = progress.values[i];
```

woodwoerk (approved):
Thanks for adding unit tests 🥇 some nitpicks but feel free to merge

PR #261, opened 2024-01-29
Remove legacy translations

woodwoerk:
No need to check individual translation strings that I've removed, but I'd appreciate a review on my general approach to make sure I haven't missed anything and removed translation strings that are in use.

How I identified translation strings to remove:

1. Performed a global search across the codebase for regex `/t\([^']/` to find calls to the translate function _without_ a string literal e.g. [here](https://github.com/kausaltech/kausal-watch-ui/pull/261/files#diff-59b1f63a4d17762600ae7d7015aa40db80dfecb696b77acf9e3449d563d6e983). I updated these to use string literals, which makes the code more readable and ensures these are picked up by TS and step 2 below.
2. Used the VSCode [i18n Ally](https://next-intl-docs.vercel.app/docs/workflows/vscode-integration#i18n-ally) extension to identify translation strings that aren't used
3. Double checked any unused strings picked up by i18n Ally weren't in use with a global search.

The regex search in step 1 would miss the following cases, but I'm quite confident we're not doing this:
1. If we've reassigned the `t` function to another variable
2. If we're using string `t('foo-' + bar)` string concatenation


terotik (commented):
Thanks for the clean-up! Seems all-ok.

tituomin (approved):

PR #262, opened 2024-01-29
Fix/accessibility html errors

lilia1891:
1. Fixed Axe-core accessibility issue: "ul must only directly contain li". LanguageSelector is an element of ul but was rendered as div. Wrapped it in ListItem to resolve the issue.
 2. Axe detected the issue with the missing title attribute for iframe (Plotly embeds). This doesn't affect user experience (screenreaders read the title of the embed anyway) and it seems we don't have the ability to manipulate this HTML before it's rendered. So I decided to exclude that specific violation from e2e accessibility testing.

### DIFF
--- a/components/common/LanguageSelector.tsx
+++ b/components/common/LanguageSelector.tsx
@@ -13,6 +13,22 @@ import { PlanContextFragment } from '@/common/__generated__/graphql';
 import { usePlan } from '@/context/plan';
 import { useApolloClient } from '@apollo/client';
 
+const LanguageSelectorListItem = styled.li`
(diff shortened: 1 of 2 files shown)


terotik:
I wonder if the language selector changes have conflicts with the changes I made in https://github.com/kausaltech/kausal-watch-ui/pull/263 ?

lilia1891:
> I wonder if the language selector changes have conflicts with the changes I made in #263 ?

It had. I resolved the conflicts

terotik (approved):

PR #269, opened 2024-02-01
Added site name to openGraph for action page

lilia1891:
---
  - https://app.asana.com/0/0/1206492816208045


app/[domain]/[lang]/[plan]/(with-layout-elements)/actions/[id]/page.tsx
```diff
@@ -55,6 +65,7 @@ export async function generateMetadata(
       description: data.action.name,
       images: image?.social?.src ? [image.social.src] : undefined,
       url: resolvedParent.openGraph?.url ?? undefined,
+      siteName,
```
Comment (woodwoerk) [inline app/[domain]/[lang]/[plan]/(with-layout-elements)/actions/[id]/page.tsx:68]: Nicely done, but we can simplify this by using the parent route's metadata. `parent` is automatically passed to to `generateMetadata` as a promise which includes all parent metadata. See `url` above for an example of this.

Comment (lilia1891) [reply app/[domain]/[lang]/[plan]/(with-layout-elements)/actions/[id]/page.tsx]: Cool :) Thanks, done

PR #270, opened 2024-02-01
Fix missing action term context for i18n string see-all-actions

bbliem:
Magdalena noticed a string is not displayed properly: https://kausaltech.slack.com/archives/C038B91QJ10/p1706785499858319?thread_ts=1706623936.596349&cid=C038B91QJ10


bbliem:
I just copied some code we have in other places. Not sure if this is the best way. The type checker complains about some things, but it also does so where I copied it from.

woodwoerk (approved):
Looks good! I'll check if we're missing that context elsewhere

PR #272, opened 2024-02-01
Fix missing search results

woodwoerk:
When typing the same search query twice, no results were displayed for the second query.

---
  - https://app.asana.com/0/0/1206486884287691

### DIFF
--- a/common/search.ts
+++ b/common/search.ts
@@ -1,6 +1,5 @@
 import { ApolloClient, InMemoryCache, gql } from '@apollo/client';
-import { PlanType } from 'context/plan';
-import { title } from 'process';
+import { PlanContextFragment } from './__generated__/graphql';
 
 const GET_AUTOCOMPLETE_RESULTS = gql`
   query GetAutocompleteResults($plan: ID!, $term: String!) {
@@ -46,7 +45,7 @@ const GET_AUTOCOMPLETE_RESULTS = gql`
 
 class WatchSearchAPIConnector {
   apolloClient: ApolloClient<InMemoryCache>;
-  plan: PlanType;
+  plan: PlanContextFragment;
 
   constructor({ apolloClient, plan }) {
     this.apolloClient = apolloClient;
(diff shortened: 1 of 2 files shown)


components/common/NavbarSearch.js
woodwoerk, line 1: Note: Everything except the loading status changes (lines 267-284) is just refactoring. I thought the issue may have been due to using React hooks in the `WithSearch` callback so refactored that callback to a separate component. Didn't turn out to be the issue but at least React isn't complaining at us now.

common/search.ts
```diff
@@ -46,20 +45,14 @@ const GET_AUTOCOMPLETE_RESULTS = gql`
...
     this.plan = plan;
   }
 
   async onSearch(opts, queryConfig) {
-    // Not implemented, stub here just to silence the warning about
-    // unimplemented method.
-    return { results: [] };
-  }
-
-  async onAutocomplete(opts, queryConfig) {
```
woodwoerk, line 62: @terotik This was a pain to debug but it turns out the issue was an `onSearch` and `onAutocomplete` race condition. `onSearch` would resolve after `onAutocomplete` when querying searches for a second time, therefor overriding the list of results with an empty array. Any idea why we also used `onAutocomplete` here? Just using `onSearch` seems to do the trick.

terotik: I'm getting `Uncaught No onAutocomplete handler provided and no Connector provided. You must configure one or the other.` warning on console running this locally. Could that be the reason? Not sure if it's a problem if it works though...

woodwoerk: That might be a problem in production builds, I'll take another look

woodwoerk: @terotik Can you test this locally again? I disabled autocomplete [here](https://github.com/kausaltech/kausal-watch-ui/pull/272/commits/83ae1463aee0994dd386ad062d7c3cf28296d851)

terotik (approved):

PR #280, opened 2024-02-06
Fix actions with special characters in their ID

woodwoerk:
The action ID is encoded by Next, which then led to the action not being found when querying GQL. Decode:
- Action IDs
- Embed slugs


lilia1891 (approved):
Great!

PR #281, opened 2024-02-06
Add initial frontend authentication

woodwoerk:
I've initially set up authentication with GitHub auth for testing until the backend is ready. It's using a custom GitHub app under my account, to test this, you'll need the following in your local `.env`:

```env
NEXTAUTH_URL=http://sunnydale.localhost:3000
AUTH_SECRET=<any randomly generated string>
AUTH_GITHUB_ID=Iv1.31ef449d4ad4b71b
AUTH_GITHUB_SECRET=<get this from Joe>
```

Click the "Sign in" button in the footer to test.


app/[domain]/[lang]/[plan]/(with-layout-elements)/layout.tsx
```diff
@@ -20,6 +21,9 @@ type Props = {
 export default function Layout({ children, params }: Props) {
   const plan = usePlan();
   const { navigationTitle, title } = getMetaTitles(plan);
+  const session = useSession();
+
+  console.log('SESSION', session);
```
Comment (woodwoerk) [inline app/[domain]/[lang]/[plan]/(with-layout-elements)/layout.tsx:26]: @lilia1891 You'll be able to use this to check whether the user is authenticated, if `session.status === 'authenticated'` then we should show the new top bar.

config/auth.ts
```diff
@@ -0,0 +1,36 @@
...
+    providers: [
+      GitHub({ redirectProxyUrl: url }),
+      {
+        id: 'watch-oidc-provider',
+        name: 'Kausal Watch Provider',
+        type: 'oidc',
+        issuer: 'https://my.oidc-provider.com',
+        redirectProxyUrl: url,
+        clientId: process.env.CLIENT_ID,
+        clientSecret: process.env.CLIENT_SECRET,
```
woodwoerk, line 30: @tituomin This is the config for our custom OIDC provider

components/Footer.tsx
```diff
@@ -18,57 +22,55 @@ const getFeedbackUrl = (currentURL) => {
...
-
-  if (plan.staticPages) {
-    const topMenuPages = plan.staticPages.filter((page) => page.footer);
-    staticPages = topMenuPages.map((page, index) => ({
-      id: `s${index}`,
-      name: page.name,
-      slug: page.slug,
-    }));
-    navLinks = navLinks.concat(staticPages);
-  }
```
woodwoerk, line 63: @terotik @tituomin `staticPages` seemed to be legacy, I couldn't find it set anywhere or queried in GQL. Can you confirm this is alright to remove?

terotik: I would say ok to remove

PR #283, opened 2024-02-07
Add missing cache headers

woodwoerk:
The `x-cache-plan-domain` and `x-cache-plan-identifier` headers were missing from GQL queries post Next.js upgrade. 

- To support setting cache headers from the RSC Apollo client, headers are set in middleware, read upon Apollo client initialisation and set as initial context
- To support setting cache headers from the client-side Apollo client, Apollo context is updated while rendering the plan page
- `headersMiddleware` reads the domain and plan ID from Apollo context and sets the query headers

### DIFF
--- a/app/[domain]/[lang]/[plan]/UpdateApolloContext.tsx
+++ b/app/[domain]/[lang]/[plan]/UpdateApolloContext.tsx
@@ -1,0 +1,21 @@
+'use client';
+
+import { PropsWithChildren } from 'react';
+import { useApolloClient } from '@apollo/client';
+import { usePlan } from '@/context/plan';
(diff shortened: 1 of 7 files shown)


woodwoerk:
Example headers set from localhost below, `X-Cache-Plan-Domain` corresponds to the hostname e.g. `sunnydale.test.kausal.tech` in testing. @juyrjola does this header need to be a specific value or is it just for looking up the cache?

```
X-Cache-Plan-Domain:          sunnydale.localhost
X-Cache-Plan-Identifier:        sunnydale
```

tituomin (approved):

PR #287, opened 2024-02-13
Set margin for languageSelector on mobile

lilia1891:
Minor visual bug: LanguageSelector slightly breaks the layout on mobile. Set the margin.
Before:
Screen Shot 2024-02-13 at 13 15 35
After:
Screen Shot 2024-02-13 at 13 33 40


woodwoerk (approved):
Nice catch! Might be a minor thing but these all give a feeling of quality to our end users 💪

PR #296, opened 2024-02-21
Support lowercase locales in the pathname

woodwoerk:
Redirect visitors to the correct pathname with properly cased locale (e.g. `es-US`) even if they have an incorrectly cased locale in their original URL (e.g. `es-us`)

User need:

> We are having an issue that if someone does not capitalize the “US” in this version of our url: www.city.gov/es-US AND they have never navigated to the Spanish version of the page before, they get an error. 
> 
> Is it possible to add a version of the URL: www.city.gov/es-us. With the “us” lower case so if it is typed in that way, it will still work?

### DIFF
--- a/utils/__tests__/middleware.test.ts
+++ b/utils/__tests__/middleware.test.ts
@@ -1,0 +1,64 @@
+import { PlanFromPlansQuery, getParsedLocale } from '../middleware.utils';
+
+const primaryLanguage = 'en-US';
+const otherLanguages = ['es-US', 'DOTHRAKI'];
+const MOCK_PLAN = {
+  id: 'foo',
+  identifier: 'foo',
+  otherLanguages,
+  primaryLanguage,
+} as PlanFromPlansQuery;
+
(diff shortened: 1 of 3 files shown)


utils/__tests__/middleware.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    expect(getParsedLocale(['en-us'], MOCK_PLAN)).toMatchObject({
+      parsedLocale: primaryLanguage,
+      isCaseInvalid: true,
+    });
+    expect(getParsedLocale(['plan', 'es-us', 'foo'], MOCK_PLAN)).toMatchObject({
+      parsedLocale: 'es-US',
+      isCaseInvalid: true,
+    });
+    expect(getParsedLocale(['dothraki', 'foo'], MOCK_PLAN)).toMatchObject({
+      parsedLocale: 'DOTHRAKI',
```
terotik, line 60: 🐴

PR #299, opened 2024-02-23
Migrate all env variables to runtime variables

woodwoerk:
This change allows us to use runtime variables in both server and client side code (replacing the need for NEXT_PUBLIC_ build time env variables) so that single Docker images can be deployed to multiple environments.

To support these variables client side, variables are injected into meta tags in the root server rendered layout component in the format `<meta name="env" content="KAUSAL_PUBLIC_WHATEVER=whatever"} />`

These new `KAUSAL_PUBLIC_` env variables need to be added to all environments before merging and deploying.


tituomin (approved):

terotik:
@woodwoerk this PR has an anniversary soon! Is it still relevant?

PR #309, opened 2024-03-12
Add mini version of action dependencies for action cards

woodwoerk:
Note, the icon in the screenshot is specific to San Diego. This should be demoed to and test by San Diego on Monday, I'm a bit concerned about the usability of such heavy popup tooltips.

image


lilia1891 (approved):
look good to me

PR #310, opened 2024-03-13
Make restricted fields more subtle

woodwoerk:
This is a minor change discussed a few weeks back that I just remembered. Also works well on a coloured background.


lilia1891 (approved):
Nice!

PR #317, opened 2024-03-21
Replace completed state for continuous actions

terotik:
Replace completed phase with "Continuous" in the UI if `action.scheduleContinuous===true`

Screenshot 2024-04-09 at 16 29 16

### DIFF
--- a/components/actions/ActionContent.tsx
+++ b/components/actions/ActionContent.tsx
@@ -404,7 +404,10 @@ function ActionContentProgressContainer({
 
       {!!action.implementationPhase && (
         <StyledProgressCard>
-          <PhaseTimeline activePhase={action.implementationPhase} />
+          <PhaseTimeline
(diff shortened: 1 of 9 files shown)


woodwoerk (approved):
Nice work! So am I understanding the Admin experience correctly that to display this, an action must be marked as continuous and the phase set to completed? I wonder if we could explain/ prompt this somewhere in the admin UI

terotik:
> Nice work! So am I understanding the Admin experience correctly that to display this, an action must be marked as continuous and the phase set to completed? I wonder if we could explain/ prompt this somewhere in the admin UI

Yep. Completed is also replaced by continuous on actions that are not completed yet ("Continuous incomplete"-example). There will be illogical combinations as the change is mainly cosmetic in UI. Improving the admin experience could be a follow up task.

PR #318, opened 2024-03-22
Revert some changes made within the last months to the action dashboard status donut

tituomin:
1. Revert back to counting undefined statuses in all numbers. (They are a "bad" state assuming that a plan wants to use statuses in the first place. This will incentivize adding status. If a plan doesn't want to use statuses, the donut can be hidden).
2. Also show the percentage of undefined statuses again.

(unrelated) fix the clipping of the hover text elements in the plotly graph when the mouse is near the bottom of the graph

### DIFF
--- a/common/preprocess.ts
+++ b/common/preprocess.ts
@@ -86,15 +86,12 @@ const getStatusData = (
   actions: ActionListAction[],
   actionStatusSummaries: ActionStatusSummary[],
   theme: Theme,
-  unknownLabelText: string = '',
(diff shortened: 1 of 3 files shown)


tituomin:
This change is made a bit "softer" by a recent change where "not started" actions are counted separately (as "good" unless they are marked late, blocked, etc.). This would make it easier for new plans with lots of not started actions.

woodwoerk (approved):

PR #324, opened 2024-03-26
Fix/a11y contrasts

terotik:


components/common/ActionAttribute.tsx
```diff
@@ -17,30 +17,12 @@ import {
...
-  }
-
-  &.icon-off {
-    fill: ${(props) => props.theme.themeColors.light} !important;
-  }
-
-  &.icon-bad {
-    fill: ${(props) => props.theme.graphColors.red070} !important;
-  }
+const ScaleIcon = styled(Icon)<{ $active?: 'sm' | 'md' }>`
```
woodwoerk, line 20: Looks like the old type is sneaking in 🥷 

```suggestion
const ScaleIcon = styled(Icon)<{ $active?: boolean }>`
```

components/common/Card.tsx
```diff
@@ -3,22 +3,30 @@ import { Card as BSCard, CardBody } from 'reactstrap';
...
   width: 100%;
   transition: all 0.5s ease;
   overflow: hidden;
   border-width: ${(props) => props.theme.cardBorderWidth};
   border-radius: ${(props) => props.theme.cardBorderRadius};
   background-color: ${(props) =>
-    props.customcolor ? props.customcolor : props.theme.themeColors.white};
+    props.$customBackgroundColor
+      ? props.$customBackgroundColor
+      : props.theme.themeColors.white};
```
woodwoerk, line 18: Little nitpick for the ternary statements in this component:

```suggestion
    props.$customBackgroundColor || props.theme.themeColors.white};
```

components/common/Card.tsx
```diff
@@ -146,8 +156,9 @@ const Card = (props: CardProps) => {
 
   return (
     <StyledCard
-      className={`${negative && 'negative'} ${outline && 'outline'}`}
-      customcolor={customColor}
+      className={`${outline && 'outline'}`}
```
woodwoerk, line 159: Nitpick: This will apply `class="false"` to the DOM element

```suggestion
      className={outline && 'outline'}
```

woodwoerk (approved):
Nice work, quite a beast of a PR! Should we test this systematically somehow or are the new Storybook additions enough (that ThemeAccessibilityTest looks like it'll come in handy)

components/common/ActionAttribute.tsx
terotik: Yes! Thanks

PR #325, opened 2024-03-27
Add cron schedule for nightly e2e test

lilia1891:
Implement nightly scheduled run of e2e tests via GitHub Actions workflow (01:00 EEST)

### DIFF
--- a/.github/workflows/ci-e2e.yaml
+++ b/.github/workflows/ci-e2e.yaml
@@ -14,8 +14,8 @@ on:
         description: 'Base url for backend API'
         required: true
         default: 'https://api.watch.kausal.tech/v1'
-      schedule:
-        - cron: '0 22 * * *' # 22:00 UTC, 01:00 EEST
+  schedule:
+    - cron: '0 22 * * 1-5' # 22:00 UTC, 01:00 EEST
 
 jobs:
   e2e_test:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-e2e.yaml
```diff
@@ -14,6 +14,8 @@ on:
         description: 'Base url for backend API'
         required: true
         default: 'https://api.watch.kausal.tech/v1'
+      schedule:
+        - cron: '0 22 * * *' # 22:00 UTC, 01:00 EEST
```
woodwoerk, line 18: I'm definitely no expert on GH action scripts, but I _think_ `schedule` needs to be at the same level as `workflow_dispatch` since they're both separate events, i.e.

```yaml
name: E2e-tests
on:
  schedule:
    - cron: '0 22 * * *' # 22:00 UTC, 01:00 EEST
  workflow_dispatch:
```

woodwoerk, line 18: Do we need to run this on weekends? Maybe it could just be Monday to Friday

```suggestion
      schedule:
        - cron: '0 22 * * MON-FRI' # 22:00 UTC, 01:00 EEST
```

lilia1891: Good point!

PR #328, opened 2024-03-27
Reference field for indicator

Bhavatu:
Show reference under indicator graph.

Depends on this: https://github.com/kausaltech/kausal-watch-private/pull/158

Related asana: https://app.asana.com/0/1206017643443542/1206649411944555

### DIFF
--- a/components/indicators/IndicatorVisualisation.js
+++ b/components/indicators/IndicatorVisualisation.js
@@ -17,6 +17,7 @@ import GraphAsTable from 'components/graphs/GraphAsTable';
 import IndicatorGraph from 'components/graphs/IndicatorGraph';
 import { useLocale, useTranslations } from 'next-intl';
 import { captureMessage } from '@sentry/nextjs';
+import RichText from 'components/common/RichText';
 
 const GET_INDICATOR_GRAPH_DATA = gql`
   query IndicatorGraphData($id: ID, $plan: ID) {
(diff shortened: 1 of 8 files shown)


Bhavatu:
I have a feeling this is not the preferred way to do it with just a `<p>` tag, opinions appreciated!

terotik:
> I have a feeling this is not the preferred way to do it with just a `<p>` tag, opinions appreciated!

Yeah, I think that would have visual implications ie. extra vertical padding. Not sure what would be a best way to visually test this locally without the backend dependency? I think a screenshot might suffice at this point and we can have a separate front end task (with Storybook) to make it look pretty. 

Also I would imagine that reference would benefit from optional link to the source?

Bhavatu:
> > I have a feeling this is not the preferred way to do it with just a `<p>` tag, opinions appreciated!
> 
> Yeah, I think that would have visual implications ie. extra vertical padding. Not sure what would be a best way to visually test this locally without the backend dependency? I think a screenshot might suffice at this point and we can have a separate front end task (with Storybook) to make it look pretty.
> 
> Also I would imagine that reference would benefit from optional link to the source?

image

This is how it looks like now. I can also add a field for a link if that is desired.

Bhavatu:
image

As discussed with @terotik, Added the reference as a RichTextField instead to support embedding links, limiting features to links only and max_length=255

terotik (changes requested):
@Bhavatu could you add the translations still for supported languages? Machine translations should be fine for this one-word. Or you can ask for a sanity check on #help-translations , thanks!

terotik (approved):

PR #332, opened 2024-04-03
Improve hierarchical option select dropdown design

terotik:


components/common/SelectDropdown.tsx
```diff
@@ -190,9 +213,29 @@ const MultiValue = (props: MultiValueProps) => {
...
 };
 
+const Option = (props: OptionProps) => {
+  const { data, children, options } = props;
+  const { indent } = data;
+  const isHierarchical: boolean = options.some((o) => o?.indent > 0);
+  const indents: JSX.Element[] = [];
+  for (let i = 0; i < indent; i++) {
+    indents.push(<DropdownIndent key={`indent-${i}`} />);
+  }
```
woodwoerk, line 223: Is there a way to do this without mutating the props?

woodwoerk (approved):
Looks much better

PR #347, opened 2024-04-16
Add conditional visibility of feedback link in Footer

lilia1891:
Ludwigsburg requests to remove the feedback link from the Footer. 
Added conditional logic to control visibility of the link.

Related task in Asana https://app.asana.com/0/1206017643443542/1207019355477556/f
Corresponding PR https://github.com/kausaltech/kausal-themes/pull/29

### DIFF
--- a/components/Footer.tsx
+++ b/components/Footer.tsx
@@ -111,18 +111,16 @@ function Footer() {
     utilityLinks.push({ id: '1', name: t('contact'), slug: plan.contactLink });
   }
 
-  if (theme.showFeedbackLink !== false) {
-    if (plan.externalFeedbackUrl) {
-      utilityLinks.push({
-        id: '2',
-        name: t('give-feedback'),
-        slug: plan.externalFeedbackUrl,
-      });
-    } else if (pathname !== '/feedback') {
-      const url = getFeedbackUrl(pathname);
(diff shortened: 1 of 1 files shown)


components/Footer.tsx
```diff
@@ -111,16 +111,18 @@ function Footer() {
...
-    if (url != null) {
-      utilityLinks.push({ id: '2', name: t('give-feedback'), slug: url });
+  if (theme.showFeedbackLink !== false) {
+    if (plan.externalFeedbackUrl) {
+      utilityLinks.push({
+        id: '2',
+        name: t('give-feedback'),
+        slug: plan.externalFeedbackUrl,
+      });
+    } else if (pathname !== '/feedback') {
```
terotik, line 121: I would probably put the `showFeedbackLink` check here. If someone has set `externalFeedbackUrl` in the admin we can safely assume that that should be an override. 

Would something like `(pathname !== '/feedback' && theme.showFeedbackLink !== false)` work?

terotik (changes requested):

components/Footer.tsx
lilia1891: Thanks @terotik! fixed the logic

PR #348, opened 2024-04-16
Support invalidating Apollo middleware cache

woodwoerk:
---
  - https://app.asana.com/0/0/1207146661131041

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -59,6 +59,20 @@ export const config = {
   ],
 };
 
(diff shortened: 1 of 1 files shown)


tituomin (approved):
I'm ok with it, if you are. But we're missing the logging for the _invalidate-middleware-cache endpoint still?

woodwoerk:
> But we're missing the logging for the _invalidate-middleware-cache endpoint still?

Yeah we are, I'll add that before merging

PR #349, opened 2024-04-17
Display action description block field label if one is defined

tituomin:
Take the field label from the block configuration of
the action page.


lilia1891 (approved):

lilia1891:
Great! Thank you, Timo!

PR #350, opened 2024-04-18
Append locale to URL when navigating from relative links

lilia1891:
A bug occurs in a multiplan with language selection: when navigating from child plans (where a language other than the primary is selected) to the parent plan via the home link, the locale is dropped from the URL, causing users to revert to the primary language.

Appended the locale to the absolute URL path when navigating from relative links.

Related Asana task https://app.asana.com/0/1203773030383185/1206977267986284/f

### DIFF
--- a/components/common/GlobalNav.tsx
+++ b/components/common/GlobalNav.tsx
@@ -26,6 +26,7 @@ import NavbarSearch from './NavbarSearch';
 import { usePlan } from 'context/plan';
 import { isServer } from 'common/environment';
 import { useTranslations } from 'next-intl';
+import { useLocale } from 'next-intl';
 
 const baseFixedNavStyles = css`
   @keyframes slide-in {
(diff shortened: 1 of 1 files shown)


components/common/GlobalNav.tsx
```diff
@@ -26,6 +26,7 @@ import NavbarSearch from './NavbarSearch';
 import { usePlan } from 'context/plan';
 import { isServer } from 'common/environment';
 import { useTranslations } from 'next-intl';
+import { useLocale } from 'next-intl';
```
terotik, line 29: Maybe this could be added to the previous `next-intl` import?

components/common/GlobalNav.tsx
```diff
@@ -487,7 +488,17 @@ function GlobalNav(props) {
...
+  const locale = useLocale();
+  let rootLink = '/';
+
+  if (plan.parent && plan.parent?.viewUrl) {
+    rootLink = plan.parent?.viewUrl;
+    const shouldAppendLocale = locale !== plan.primaryLanguage;
+    if (shouldAppendLocale) {
+      rootLink = `${plan.parent?.viewUrl}/${locale}/`;
+    }
+  }
```
woodwoerk, line 500: Nitpick: To improve readability, this could be moved to a dedicated `getRootLink` function. We also then wouldn't need to reassign `rootLink` but could just return the value based on the different if statements (in general we prefer not to use `let`)

terotik (changes requested):

woodwoerk (approved):
Looks good to me! One nitpick but no need to implement it if you're working on other stuff

terotik (approved):
Great thanks!

PR #351, opened 2024-04-19
Add continuous actions to phase summary graphs

terotik:
Add completed continuous actions to phase summary graphs
Fix reversed phase color order in some phase status bars

Screenshot 2024-04-19 at 17 21 35
Screenshot 2024-04-19 at 17 21 17

### DIFF
--- a/components/contentblocks/ActionStatusGraphsBlock.tsx
+++ b/components/contentblocks/ActionStatusGraphsBlock.tsx
@@ -19,6 +19,7 @@ const GET_ACTION_LIST_FOR_GRAPHS = gql`
   query GetActionListForGraphs($plan: ID!, $categoryId: ID) {
     planActions(plan: $plan, category: $categoryId) {
       color
+      scheduleContinuous
       statusSummary {
         identifier
       }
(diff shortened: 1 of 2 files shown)


components/contentblocks/ActionStatusGraphsBlock.tsx
```diff
@@ -39,12 +40,8 @@ interface Props
...
 }
 
-const ActionStatusGraphsBlock = ({
-  id = '',
-  categoryId,
-  columnProps,
-  ...graphsProps
-}: Props) => {
+const ActionStatusGraphsBlock = (props: Props) => {
+  const { id = '', categoryId, columnProps, ...graphsProps } = props;
```
terotik, line 44: I prefer doing this as it often helps debugging (logging all props)

tituomin (approved):

components/contentblocks/ActionStatusGraphsBlock.tsx
woodwoerk: You could just log `graphsProps` or add `...rest` in cases where we don't already destructure the remaining props. I find this makes components harder to refactor because it's harder to spot if a prop is used or not

PR #352, opened 2024-04-22
Make minor improvements regarding accessibility

bbliem:
Nothing exciting, just stuff I noticed during my BITV accessibility testing.
- `<p>` elements were used for things that are clearly not paragraphs; replaced by `<div>`
- Sometimes `<th>` elements were nested in `<th>` elements. I un-nested them.
- Some column headers didn't have `scope="col"` set.

I also noticed that in the table on the indicator list page, there are no row headers. But I was too lazy to fix that now.

### DIFF
--- a/components/actions/ContactPerson.js
+++ b/components/actions/ContactPerson.js
@@ -36,13 +36,13 @@ const PersonDetails = styled.div`
   }
 `;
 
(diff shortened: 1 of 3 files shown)


woodwoerk (approved):
Thanks for these!

PR #358, opened 2024-04-26
Show action.status in places where that is explicitly desired

tituomin:
The action dashboard status column and the action details page should show just the plain action status in locations designed to show it.

Action.statusSummary should be reserved for places with no room to separate status/phase, like ActionCards, and also the action tables in indicator pages.


woodwoerk (approved):
I haven't tested this locally but looks like it'll do the trick!

PR #360, opened 2024-05-01
Fix links to cross-plan merged actions

bbliem:
[Asana task](https://app.asana.com/0/1206017643443542/1207112938358823/f)

Unfortunately this creates links to whatever the view URL of an action is, so in a development deployment it will not point to localhost. But this seemed to be the case already before for cross-plan action links in general.


woodwoerk (approved):
LGTM!

PR #365, opened 2024-05-08
Add max width to avoid image overflow on mobile

woodwoerk:
Spotted by Samuel [here](https://climatedashboard.durham.ca/actions/A2)

| Before | After |
| -- | -- |
| durham-ca-cap localhost_3000_actions_A2(iPhone SE) (3) | durham-ca-cap localhost_3000_actions_A2(iPhone SE) (1) |


terotik:
> @terotik this _feels_ like an innocent change, but can you take a moment to think of any possible side effects?

Can't think of any. We do not want images to stretch the page on mobile widths. I kind of assumed this was already in place - maybe it's some other image instance I'm thinking of. Good fix nevertheless, thanks!

PR #373, opened 2024-05-22
Sort Action Related Indicators by name

Bhavatu:
Sort Action Related Indicators by name.

Asana: https://app.asana.com/0/1206017643443542/1207297197056775


Bhavatu:
Not sure if this has any unwanted implications. If so, a custom ordering could be implemented instead.

PR #375, opened 2024-05-31
Fix/accessibility improvment

lilia1891:
Accessibility improvements across elements (missing roles, attributes) based on Superior report.
Related Asana task and the report: https://app.asana.com/0/1205310117865974/1207079524711057/f

### DIFF
--- a/components/actions/ActionCard.tsx
+++ b/components/actions/ActionCard.tsx
@@ -419,7 +419,12 @@ function ActionCard({
             <StyledActionDependencyIconWrapper
               id={getDependencyTooltipId(action.id)}
             >
-              <Icon name="action-dependency" width="24px" height="24px" />
+              <Icon
+                name="action-dependency"
+                width="24px"
+                height="24px"
+                role="presentation"
+              />
             </StyledActionDependencyIconWrapper>
             <StyledTooltip
               target={getDependencyTooltipId(action.id)}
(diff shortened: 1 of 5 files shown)


components/actions/ActionCard.tsx
terotik, line 1: Should we announce this action dependency to screenreaders? @woodwoerk . Maybe we have some translation string already in place that could be used as title?

components/common/Card.tsx
```diff
@@ -130,6 +132,8 @@ const Card = (props: CardProps) => {
             $background={imageUrl}
             $imageAlign={imageAlign}
             data-testid="image-bg"
+            role="img"
+            aria-label={t('image-credit')}
```
terotik, line 136: Maybe this should be `image. altText` (although it's not always filled, but then the responsibility is on content editor)?

terotik (changes requested):
I added some comments

components/actions/ActionCard.tsx
woodwoerk: Yeah, let's do that in a separate PR, it doesn't need to block this. I'll add a task!

components/common/Card.tsx
lilia1891: @terotik I added a condition to use ariaText for aria-label if it's provided. Otherwise, we use generic "image"

lilia1891: Also, maybe we should make the aria-text field for images required in the Admin to avoid empty aria-labels for images?

terotik (approved):

PR #376, opened 2024-06-03
Automatically logout when the session has expired

tituomin:
Prevents the UI ending up in a logged in state but unable to actually retrieve restricted data.


woodwoerk (approved):
Noice!

PR #377, opened 2024-06-05
Hide avatar based on plan feature setting

Bhavatu:
Hide avatars if plan feature setting requires it.
To signal "leader" we show a left border instead of the circle around the avatar.

image

Depends on:
Backend: https://github.com/kausaltech/kausal-watch-private/pull/206

Asana: https://app.asana.com/0/1206017643443542/1207437856881925

### DIFF
--- a/components/actions/ContactPerson.js
+++ b/components/actions/ContactPerson.js
@@ -25,6 +25,10 @@ const Person = styled.div`
       border: 4px solid ${(props) => props.theme.brandDark};
     }
   }
+
+  &.leader-no-avatar {
+    border-left: 5px solid ${(props) => props.theme.brandDark};
+  }
 `;
 
 const PersonDetails = styled.div`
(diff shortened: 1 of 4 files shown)


components/actions/ContactPerson.js
```diff
@@ -25,6 +25,10 @@ const Person = styled.div`
       border: 4px solid ${(props) => props.theme.brandDark};
     }
   }
+
+  &.leader-no-avatar {
+    border-left: 5px solid ${(props) => props.theme.brandDark};
+  }
```
woodwoerk, line 31: The old code goes against our conventions here, so this might be a good time to refactor. Rather than using css classes, we should pass props to styled components to avoid any global classes which accidentally pick up other styles. Styled components will automatically create scoped classnames for us and makes sure they're server side rendered.

That would look something like this (I haven't tested it):

```ts
const Person = styled.div<{ $isLeader?: boolean, $withoutAvatar?: boolean }>`
  display: flex;
  margin-top: 1em;
  padding-bottom: 1em;
  border-bottom: 2px solid ${(props) => props.theme.themeColors.light};
  img {
    border: 2px solid ${(props) => props.theme.themeColors.light};
  }

  ${({ $isLeader }) => $isLeader && `
    img {
      border: 4px solid ${(props) => props.theme.brandDark};
    }
  `}

  ${({ $isLeader, $withoutAvatar }) => $withoutAvatar && $isLeader && `
    border-left: 5px solid ${(props) => props.theme.brandDark};
  `}}
```

Then the styles can be simply controlled via props: `<Person $isLeader={leader} $withoutAvatar={...}>...`

woodwoerk (approved):
Nice!

PR #380, opened 2024-06-13
Become compatible with the new theme builds

tituomin:
Theme theme infra is changing between major versions 0.4 and 0.5,
and we need these changes to make the UI compatible

Review instructions:
The latter commit with the Icon usage changes is not strictly necessary, but I find it nice and it should only contain changes of the exact same type, while the former commit contains the actual real changes.

For the latter commit, notice that the old style name="xxx" still works too and is used in some places where it's more convenient. If the change feels unnecessary, we can just remove it and keep the old API.

### DIFF
--- a/common/theme.ts
+++ b/common/theme.ts
@@ -25,6 +25,6 @@ export async function loadTheme(themeIdentifier: string): Promise<Theme> {
   }
 }
 
-export function getThemeCSS(themeIdentifier: string) {
-  return `/static/themes/${themeIdentifier}/main.css`;
+export function getThemeStaticURL(path: string) {
+  return `/static/themes/${path}`;
 }
--- a/app/[domain]/[lang]/[plan]/embed/v1/[...slug]/page.tsx
+++ b/app/[domain]/[lang]/[plan]/embed/v1/[...slug]/page.tsx
@@ -29,8 +29,7 @@ const UnknownEmbedPlaceholder = () => {
   const theme = useTheme();
   return (
     <UnknownWrapper>
-      <Icon
-        name="exclamation-circle"
+      <Icon.ExclamationCircle
         color={theme.themeColors.light}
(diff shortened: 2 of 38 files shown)


context/shared-icons.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const iconReferences = Object.values(AvailableIcons);
+  const hiddenIcons = iconReferences.map((iconRef) => (
+    <HiddenReusableIcon key={iconRef} name={iconRef} />
+  ));
+
+  return (
+    <SharedIconContext.Provider value={{ iconReferences }}>
+      <div style={{ display: 'none' }}>{hiddenIcons}</div>
+      {props.children}
+    </SharedIconContext.Provider>
```
woodwoerk, line 25: Is context necessary here? Would it be enough to just render the `hiddenIcons`

woodwoerk (approved):
Really nicely done 👏

common/theme.ts
```diff
@@ -25,6 +25,17 @@ export async function loadTheme(themeIdentifier: string): Promise<Theme> {
   }
 }
 
-export function getThemeCSS(themeIdentifier: string) {
-  return `/static/themes/${themeIdentifier}/main.css`;
+const assetPrefix = ''; // FIXME: take from environment variable
```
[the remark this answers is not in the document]
tituomin: Right, I don't actually know, in paths it's an env variable, we might as well add this only when it is needed.

context/shared-icons.tsx
tituomin: No, there was an earlier, too complicated version where the context was actually updated based on which icons are actually used, but I think that's unnecessary. Will simplify.

tituomin:
@juyrjola Yes, I think so. Though I wanted to double check the favicons work

PR #381, opened 2024-06-28
Add columns for custom fields

terotik:


components/dashboard/dashboard.constants.tsx
```diff
@@ -130,4 +181,47 @@ export const COLUMN_CONFIG: { [key in ColumnBlock]: Column } = {
...
+        (a) => a.type.id === attributeType.id
+      );
+      if (!attributeContent) return null;
+      return (
+        <ActionAttribute
+          attribute={attributeContent}
+          attributeType={attributeType}
+          notitle
+          minimized
+        /> // TODO: Render attribute content
```
tituomin, line 197: I suppose this is already done?

tituomin (commented):
The end result looks cleaner than it was before. No obvious problems, looks good! I will need to test some perf implications for this change.

PR #383, opened 2024-07-23
Escape user input before using regex

lilia1891:
* Added the library to handle the escaping;
* Updated `filterAction` method;

Thread related to the bug  https://kausaltech.slack.com/archives/C01CZMVSM9C/p1721725603710529


terotik (commented):
Looks good, I did not run it locally but trust it working

terotik (approved):

PR #387, opened 2024-07-31
Refactor: Handle absent avatars

lilia1891:
Display an empty circle when the avatar image is not provided (and not intentionally hidden).
* Replaced the `img` element with a styled `div` to get more control over avatar's placeholder;
* Added `hasAvatar` prop for conditional rendering;
* Removed the default path;
* Added `aria-label` for accessibility.

Before:
Screen Shot 2024-07-31 at 13 10 43

After:
Screen Shot 2024-08-06 at 16 27 19

### DIFF
--- a/components/actions/ContactPerson.tsx
+++ b/components/actions/ContactPerson.tsx
@@ -83,12 +83,10 @@ const Avatar = styled.div<{
   width: 5em;
   height: 5em;
   border-radius: 50%;
-  background-color: ${(props) =>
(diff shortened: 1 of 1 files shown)


terotik (changes requested):
Good refactor, but I would still suggest to keep the default fallback avatar we get from the themes, so you get this "icon" instead of the empty circle

Screenshot 2024-08-06 at 12 24 42

terotik (approved):
Looks good!

PR #392, opened 2024-08-13
Format time-to-goal better

terotik:
When we are less than a year from the indicator goal, the summary displays 0 years
Screenshot 2024-08-13 at 20 29 51

Use `dayjs().to` to make time-to-goal human readable and localised
Screenshot 2024-08-13 at 20 29 43

Add "+" to emphasise expected increase as opposed to decrease
Screenshot 2024-08-13 at 20 28 20

Also added a few quick customer requested improvements on plotly graph formatting.

### DIFF
--- a/components/graphs/IndicatorGraph.tsx
+++ b/components/graphs/IndicatorGraph.tsx
@@ -1,13 +1,13 @@
 'use client';
 
 import React from 'react';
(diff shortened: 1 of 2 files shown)


lilia1891 (approved):
Nice! Gets more informative now

PR #395, opened 2024-08-21
Custom fields support added to feedback form

Bhavatu:
Support for custom additional fields in feedback form

Depends on: https://github.com/kausaltech/kausal-watch-private/pull/247

image


components/common/CheckboxInput.tsx
```diff
@@ -0,0 +1,95 @@
...
+      (optionValue: string) => (e: React.ChangeEvent<HTMLInputElement>) => {
+        const newValue = e.target.checked
+          ? [...value, optionValue]
+          : value.filter((v) => v !== optionValue);
+
+        const event = {
+          target: { value: newValue },
+        };
+
+        onChange(event as React.ChangeEvent<HTMLInputElement>);
```
woodwoerk, line 65: I think TS will complain about this as event target values should be a string. Instead we can update `CheckboxInputProps` `onChange` to use the correct react-hook-form type.

```
  ...
  onChange: ControllerRenderProps['onChange'];
};
```

components/common/CheckboxInput.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <GroupLabel>{heading}</GroupLabel>
+        {options.map((option, index) => (
+          <Label key={option.value} htmlFor={`${id}-${index}`}>
+            <Checkbox
+              id={`${id}-${index}`}
+              type="checkbox"
+              value={option.value}
+              checked={value.includes(option.value)}
+              onChange={handleChange(option.value)}
+              innerRef={index === 0 ? ref : undefined}
```
woodwoerk, line 79: What's the purpose of adding the ref to the first checkbox?

components/common/FeedbackForm.js
```diff
@@ -73,6 +82,108 @@ const FeedbackForm = ({
...
+            name={field.fieldLabel}
+            control={control}
+            defaultValue=""
+            rules={{
+              required: field.fieldRequired && t('error-field-required'),
+            }}
+            render={({ field: controllerField }) => (
+              <TextInput
+                {...controllerField}
+                id={`dynamic-${field.fieldLabel}`}
```
woodwoerk, line 102: We should avoid using ids with white space. Typically ids should just be alphanumeric, dashes and underscores.

I can't remember what happens, but it might not explode, so this can probably be sorted later if we don't have something like `field.fieldId` available for an easy fix.

woodwoerk (approved):
Great work! Messing with React Hook Form ain't simple 😊 Added some nitpicks/ suggestions, these can be sorted later if this is urgent.

PR #406, opened 2024-09-05
Feature/navbar search query update

lilia1891:
Asana task https://app.asana.com/0/1205310117865974/1208062682611022/f

A customer request: 

* `NavbarSearch` should return the result by keywords used not only in the titles but in the content too. 
    Applied `SearchQuery` instead of `GetAutocompleteResults` to the component to expand the search.
*  Fix the bug: input field unexpectedly disappears while entering keywords
    Wrapped the `NavbarSearch` component in `React.memo` to prevent unnecessary re-renders. 

https://github.com/user-attachments/assets/755038c9-6bce-4123-898c-96a99b32120a


woodwoerk (approved):
Nice work! Nice solution for the disappearing search results box too 👍

PR #408, opened 2024-09-10
Fix some icon references

terotik:
Hopefully the last ones!


terotik:
I have a vscode extension that organises the imports on save. Not sure if great for PR readability, but I also like them in order 🍡

tituomin:
Maybe then save first, commit with message "formatting" and then do the changes.

tituomin (approved):

PR #410, opened 2024-09-11
Close result box after navigation to search page

lilia1891:
The bug was caused by the previous fix: applying `React.memo` for `NavbarSearch` component to prevent unnecessary re-renders. 
When a user enters a keyword in the navbar search, clicks on the "See all results" link and gets redirected to the Search page, the result box is still displayed, but after closing it "No results for this search..." message is displayed, even though there were search results.

https://github.com/user-attachments/assets/ddfa0d58-cc39-4101-83ed-ff2692ceb0fc

* Added `closeSearch()` in the ResultList to close the result box after the link is clicked, redirecting the user to the search page;
* Removed `setSearchTerm('')` to ensure that the search term is passed and the results are displayed on the Search page.

After:

https://github.com/user-attachments/assets/6b7f6fcb-e716-4f97-911e-a74308ea46f0

### DIFF
--- a/components/common/NavbarSearch.js
+++ b/components/common/NavbarSearch.js
@@ -249,7 +249,7 @@ function ResultItem(props) {
 }
 
 const ResultList = (props) => {
-  const { results, searchTerm, loading } = props;
+  const { results, searchTerm, loading, closeSearch } = props;
   const t = useTranslations();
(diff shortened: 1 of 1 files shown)


woodwoerk (approved):
Nice detective work 🕵️ 🐛

PR #413, opened 2024-09-27
Use new indicator visualisation options

terotik:
BEFORE we assumed the desired trend and that the goal is never reached
Screenshot 2024-09-27 at 19 01 33

NOW we set the desired trend in admin and tell if the goal has already been exceeded
Screenshot 2024-09-27 at 20 15 26

Also the trendline is shown if so set in admin

### DIFF
--- a/components/indicators/IndicatorValueSummary.js
+++ b/components/indicators/IndicatorValueSummary.js
@@ -60,6 +60,8 @@ function determineDesirableDirection(desiredTrend, values, goals) {
     return '-';
   }
 
+  // Default value for desiredTrend is `A_` that wil try to
+  // guess the desired trend based on values and goals
   if (!values.length || !goals.length) return null;
 
   const latestValue = values[values.length - 1];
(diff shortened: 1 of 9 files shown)


components/indicators/IndicatorValueSummary.js
```diff
@@ -52,7 +52,14 @@ const ChangeSymbol = styled.span`
     `${props.theme.fontFamilyTiny}, ${props.theme.fontFamilyFallback}`};
 `;
 
-function determineDesirableDirection(values, goals) {
+function determineDesirableDirection(desiredTrend, values, goals) {
```
tituomin, line 55: Since this function is only called once below, I don't see much use for it, since I think `INCREASING` is a bit more descriptive than `+`anyway.

tituomin: Especially since the `values` and `goals` are not used.

tituomin (approved):
Minor nag, but approved!

components/indicators/IndicatorValueSummary.js
terotik: The default value for desiredTrend is `A_` (Automatically detect desired trend) so basically this provides the default fall back solution if no desiredTrend is set. I shall comment it in the code as `A_` is not immediately obvious in ui code

PR #417, opened 2024-10-04
Remove link localization logic from UI

tituomin:
The backend now provides viewUrls already localized provided
based on the locale directive


woodwoerk (approved):
Nice one, looks good to me! 🍟

PR #419, opened 2024-10-15
Use relatedPlanIndicators in umbrella plan. Filter by common categories.

Bhavatu:
Use relatedPlanIndicators in umbrella plan instead of planindicators. Filter by relatedPlanIndicators common categories instead of planIndicator categories

### DIFF
--- a/components/indicators/IndicatorListFiltered.tsx
+++ b/components/indicators/IndicatorListFiltered.tsx
@@ -403,6 +403,8 @@ const IndicatorListFiltered = (props) => {
     displayNormalizedValues,
     shouldDisplayCategory,
     displayLevel,
+    includePlanRelatedIndicators,
+    commonCategories,
   } = props;
 
   const locale = useLocale();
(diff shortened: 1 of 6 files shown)


components/indicators/IndicatorListFiltered.tsx
```diff
@@ -692,19 +704,41 @@ const IndicatorListFiltered = (props) => {
...
+                                      commonCategory.typeIdentifier
+                                )
+                                .map((cat) => (
+                                  <StyledBadge key={cat.common.id}>
+                                    {cat.common.name}
+                                  </StyledBadge>
+                                ))}
+                            </IndentableTableCell>
+                          ))
+                        : someIndicatorsHaveCategories && (
```
tituomin, line 726: I find the ternary operator included within React markup makes the whole thing a bit hard to read. I wonder if we could either extract the two different variations into separate functions (components) or at the minimum initialize them into a const variable before this code.

I mean instead of:
`condition ? <blah blah blah blah blahz> : <another blah blah blah>`
we would have

```
const ComponentA  = (props) => <blah blah blah>
const ComponentB = (props) => <another blah blah>

...

condition ? <ComponentA props.../> : <ComponentB props.../>
```
for example

tituomin (commented):
One comment about inline ternary operators, otherwise my biggest question is if we should add more types especially to function signatures.

components/indicators/IndicatorListFiltered.tsx
Bhavatu: Refactored the ternary operator within the markup.
Also added some more typing.

tituomin (approved):

PR #420, opened 2024-10-18
Contact form feedback and email field settings

Bhavatu:
Feedback forms comment field ("Feedback") and email field can now be hidden, optional or required based on settings from the backend.

Depending on backend: https://github.com/kausaltech/kausal-watch-private/pull/281


components/common/FeedbackForm.js
```diff
@@ -69,10 +77,28 @@ const FeedbackForm = ({
   const onSubmit = (formData) => {
     const { name, email, comment, ...additionalResponse } = formData;
 
+    const isCommentEmpty = !comment || comment.trim() === '';
+    const areAllAdditionalFieldsEmpty = Object.values(additionalResponse).every(
+      (value) => {
+        if (Array.isArray(value)) {
+          return value.length === 0;
+        }
+        return value === '' || value === null || value === undefined;
```
tituomin, line 86: value.trim() ?

tituomin: Also, you can use `value == null` instead of `value === null || value === undefined`
Also, you should maybe move the null/undefined check first, then you can add the trim()

tituomin (approved):
Some non-obligatory style comments

PR #422, opened 2024-10-24
Fix trend line not extending to the goal year

lilia1891:
A bug: The trend line was not extending to the correct goal year in the Indicator graphs.
Asana https://app.asana.com/0/1206017643443542/1208538004462773/f

We were getting `highestGoalYear: NaN`. 
Modified `highestGoalYear` to extract and parse the last element of the `x` array.

Before:
before

After:
after

### DIFF
--- a/components/indicators/IndicatorVisualisation.js
+++ b/components/indicators/IndicatorVisualisation.js
@@ -311,8 +311,8 @@ const generateTrendTrace = (indicator, traces, goals, i18n) => {
     const highestDataYear = traces[0].y[traces[0].y.length - 1];
     const highestGoalYear = Math.max(
       ...goals.map((goal) => {
-        const goalYear = goal.x[goal.x.length - 1];
-        return goalYear ? parseInt(goalYear.split('-')[0], 10) : NaN;
+        const goalDate = goal.x[goal.x.length - 1];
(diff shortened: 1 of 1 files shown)


components/indicators/IndicatorVisualisation.js
```diff
@@ -310,8 +310,12 @@ const generateTrendTrace = (indicator, traces, goals, i18n) => {
 
     const highestDataYear = traces[0].y[traces[0].y.length - 1];
     const highestGoalYear = Math.max(
-      ...goals.map((goal) => goal.x[goal.x.length - 1])
+      ...goals.map((goal) => {
+        const goalYear = goal.x[goal.x.length - 1];
+        return goalYear ? parseInt(goalYear.split('-')[0], 10) : NaN;
```
tituomin, line 315: Instead of 
`parseInt(goalYear.split('-')[0], 10)`
you could maybe use
`(new Date(goalYear)).getFullYear()` to make the intention clearer.
Also, maybe rename goalYear variable to goalDate to be clearer.

tituomin (changes requested):

components/indicators/IndicatorVisualisation.js
lilia1891: Done! thank you for suggestion @tituomin

tituomin (approved):

PR #424, opened 2024-10-29
Highlight the late tasks in the Action Task List

lilia1891:
UI improvement: Display late tasks as late in action task list - Asana https://app.asana.com/0/1205310117865974/1207617873637474/f

* Reused `StatusBadge` component for visual consistency to mark the late tasks within the `ListGroupItem`.
* Highlighted the entire `ListGroupItem` when a task is marked as late, using the background color from the StatusBadge/late.
* Added English translations for the "Late" label.

Screen Shot 2024-10-29 at 09 20 01

To do next: Add translations for the "Late" label via Weblate to support other languages.

### DIFF
--- a/components/actions/TaskList.js
+++ b/components/actions/TaskList.js
@@ -140,6 +140,10 @@ function parseTimestamp(timestamp) {
   return dayjs(timestamp).format(timeFormat);
 }
 
+const isTaskLate = (task, completed) => {
+  return !completed && new Date(task.dueAt) < new Date();
+};
+
 const Task = (props) => {
   const t = useTranslations();
   const locale = useLocale();
@@ -154,7 +158,7 @@ const Task = (props) => {
     getDateFormat(dateFormat)
   );
 
-  const isLate = !completed && new Date(task.dueAt) < new Date();
+  const isLate = isTaskLate(task, completed);
 
   return (
     <TaskWrapper>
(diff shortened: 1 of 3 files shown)


components/actions/TaskList.js
```diff
@@ -196,11 +228,18 @@ function TaskList(props) {
   function filterTasks(state, completed) {
     return sortedTasks
       .filter((item) => item.state === state)
-      .map((item) => (
-        <ListGroupItem key={item.id} className={`state--${item.state}`}>
-          <Task task={item} theme={theme} completed={completed} />
-        </ListGroupItem>
-      ));
+      .map((item) => {
+        const isLate = !completed && new Date(item.dueAt) < new Date();
```
tituomin, line 232: Extract `isLate` into a function (see above)?
Also, just a curious peculiarity: `new Date()` is always in the browser's local timezone, but `item.dueAt` is in the plan's local timezone. I don't think we need to do anything about it though, because it always works as intended from the customer's own browsers. Only from the other side of the world it might sometimes be one day off but no big deal.

tituomin (approved):
Optional improvement with refactoring common code.

components/actions/TaskList.js
lilia1891: Renamed to isTaskLate and extracted into a reusable function.

PR #426, opened 2024-10-31
Fix/plan selector dropdown

lilia1891:
Fix: The plan selector dropdown menu is overlapped by other components
Asana - https://app.asana.com/0/1206017643443542/1208648881969349/f

* Updated DropdownMenu to use container="body" to ensure the dropdown is not affected by overflow styles

Before:
Screen Shot 2024-10-31 at 12 02 26

After:
Screen Shot 2024-10-31 at 11 51 34

### DIFF
--- a/components/common/GlobalNav.tsx
+++ b/components/common/GlobalNav.tsx
@@ -86,8 +86,6 @@ const BotNav = styled(Navbar)<{ $offsetTop?: number; $expanded: boolean }>`
 
   .container {
     flex-wrap: nowrap;
-    overflow-x: auto;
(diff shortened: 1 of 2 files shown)


tituomin (commented):
You should ask Tero about the commit https://github.com/kausaltech/kausal-watch-ui/commit/293582426c1bf404a5562ad3c7acf16e3f3d9d84 where those css lines were originally added, to make sure this doesn't break the fixes in that commit.

terotik (approved):
Yeah this was it!

PR #432, opened 2024-11-20
Add and display main goal attribute

lilia1891:
Add "main goal" attribute to the category and display it on category cards
Asana https://app.asana.com/0/1206017643443542/1208682503861856/f

Screen Shot 2024-11-20 at 09 24 31


terotik (approved):
Nice. You found the way to inline the label :)

PR #438, opened 2024-12-03
Feature/move spreadhseet export to backend

tituomin:
Needs to be deployed after https://github.com/kausaltech/kausal-watch-private/pull/301


components/dashboard/ActionStatusExport.js
```diff
@@ -9,193 +7,18 @@ import {
...
+  const url = plan.actionReportExportViewUrl;
+  const csvExportUrl = `${url}?format=csv`;
+  const excelExportUrl = `${url}?format=xlsx`;
   return (
     <UncontrolledDropdown>
       <DropdownToggle caret>{t('export')}</DropdownToggle>
       <DropdownMenu>
-        <DropdownItem onClick={() => handleExport('excel')}>Excel</DropdownItem>
-        <DropdownItem onClick={() => handleExport('csv')}>CSV</DropdownItem>
+        <DropdownItem href={excelExportUrl}>Excel</DropdownItem>
```
woodwoerk, line 20: Is `href` specifically supported by `DropdownItem`? If not, we should override the underlying element be an `<a>` using `tag="a"` (I think that's the prop).

You probably also want to use [the `download` property](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#download) here.

tituomin: export interface DropdownItemProps extends React.HTMLAttributes<HTMLElement> {
  [key: string]: any;
  active?: boolean;
  disabled?: boolean;
  divider?: boolean;
  tag?: React.ElementType;
  header?: boolean;
  cssModule?: CSSModule;
  href?: string;
  toggle?: boolean;
  text?: boolean;
}
I think href is supported!

tituomin:
```      } else if (props.href) {
        Tag = 'a';
      } else if (text) {
```

woodwoerk: Great! Well found, they don't mention it [in the docs](https://reactstrap.github.io/?path=/docs/components-dropdown--dropdown)

woodwoerk: If we want to be on the safe side, we could continue using `file-saver` but simplify that original `handleExport` to just download the specified URL.

```ts
  import { saveAs } from 'file-saver';
  ...
  const handleExport = async (downloadUrl) => {
    saveAs(downloadUrl);
  };
```

I've not tested that, but it's probably safer than letting browsers decide whether to allow our cross-origin download (I should do the same for NetZeroPlanner)

PR #439, opened 2024-12-05
Display image credit in LargeImageBlock

lilia1891:
Fix: The image credit for images in the Steamfield `LargeImageBlock` was not displayed in the UI.
Asana - https://app.asana.com/0/1206017643443542/1208839363867554/f

* Updated the `STREAM_FIELD_FRAGMENT` to include the `imageCredit` field for `LargeImageBlock`;
* Re-used the `ImageCredit` from `ContentPageHeaderBlock`;
* Added conditional rendering of `ImageCredit` to display it only when the backend provides the data. 

Test case: http://klima-potsdam.localhost:3000/aktivitaeten/klimapreis

### DIFF
--- a/components/common/StreamField.tsx
+++ b/components/common/StreamField.tsx
@@ -22,6 +22,8 @@ import IndicatorShowcaseBlock from 'components/contentblocks/IndicatorShowcaseBl
 import QuestionAnswerBlock from 'components/contentblocks/QuestionAnswerBlock';
 import RelatedIndicatorsBlock from 'components/contentblocks/RelatedIndicatorsBlock';
 import RelatedPlanListBlock from 'components/contentblocks/RelatedPlanListBlock';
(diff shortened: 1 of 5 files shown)


components/common/StreamField.tsx
```diff
@@ -274,16 +278,30 @@ function StreamFieldBlock(props: StreamFieldBlockProps) {
...
+              }}
             >
-              <img
-                src={block.image?.renditionUncropped?.src}
-                alt={block.image?.altText}
+              <div
                 style={{
-                  display: 'block',
-                  width: '100%',
-                  marginBottom: theme.spaces.s600,
```
woodwoerk, line 284: Should we keep this `marginBottom`  on the new `div` element?

lilia1891: We should! thanks @woodwoerk !

woodwoerk (approved):
Nice! Looks good to me

PR #440, opened 2024-12-05
Remove ImpactColumnBlock

tituomin:
This has to be merged and deployed first, only then can the backend without the ImpactColumnBlock be deployed.


woodwoerk (approved):
It's what I always wanted 🎁 🎅

PR #442, opened 2024-12-12
Resolve SectionHeader contrast issue in IndicatorGroupBlock

lilia1891:
Bug: `SectionHeader` text color in the `IndicatorGroupBlock` is not readable in some cases, due to the `readableColor` logic, which caused the text color to be overridden if color contrast insufficient.
 The bug appears after applying SectionHeader to the Indicators block for visual consistency with the ActionListBlock. 
 
 * Removed `readableColor` logic from h2.
 
 Asana https://app.asana.com/0/1206017643443542/1208963765257879/f


woodwoerk (approved):
Tested a few plans and it seems to be fine 🎄

PR #443, opened 2024-12-13
Feat/make section headers consistent

lilia1891:
Make section headers consistent across the pages - Asana https://app.asana.com/0/1206017643443542/1208848243610566/f

Initially, the idea was to make ActionListblock and IndicatorListBlock headers consistent - PR https://github.com/kausaltech/kausal-watch-ui/pull/435;
When it was decided to apply the same header style to other displayed on the same page blocks headers:
CategoryListBlock - (re-used SectionHeader from ActionListblock);
IndicatorHighlightsBlock - (applied the same styles to the header to prevent hydration errors);
ActionsHighlightBlock - (applied the same styles to the header to prevent hydration errors).

### DIFF
--- a/components/actions/ActionHighlightsList.tsx
+++ b/components/actions/ActionHighlightsList.tsx
@@ -74,8 +74,16 @@ export const GET_ACTION_LIST = gql`
 
 const ListHeader = styled(Col)`
   h2 {
+    text-align: center;
+    padding: ${(props) => props.theme.spaces.s100};
+    border-radius: ${(props) => props.theme.cardBorderRadius};
+    background-color: ${(props) => props.theme.themeColors.white};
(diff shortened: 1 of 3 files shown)


woodwoerk:
The improvement in consistency across headers is great, but I'd like to get @terotik's input on this from a visual perspective. I feel that using these header backgrounds for all cities on all pages is quite a big change, so we should be sure that:

1. We're happy with this from a visual point of view
2. We're confident that other cities won't ask us to get rid of these new header backgrounds after one city requested them 😄 (deja vu?)

My thoughts on the first point: These block background headers work well on the category pages (where we're just listing actions, indicators etc), but on the home page and other content pages may be overpowering. I'll pass the final visual decision on to @terotik.

### Examples:

image

image

terotik (approved):

PR #445, opened 2024-12-30
Another fix URL construction for LogoLink

Bhavatu:
followup to: https://github.com/kausaltech/kausal-watch-ui/pull/441/files
the first fix did not work because the viewurl included `/<locale>` and not `/<locale>/ `


lilia1891 (approved):
Great, thanks!

PR #446, opened 2024-12-30
Handle button color exception on actionlist filters

terotik:
Handle situations where action list filters background color is the same as primary button color:
Screenshot 2024-12-30 at 15 43 59

By using secondary button in those cases
Screenshot 2024-12-30 at 15 43 28


Bhavatu (approved):
LGTM!

PR #447, opened 2025-01-03
Do not block builds completing because of Sentry unavailable

tituomin:
See https://www.npmjs.com/package/@sentry/webpack-plugin#errorhandler for documentation.


Bhavatu (approved):
Lgtm!

PR #448, opened 2025-01-07
Adjust filter label text on indicators page

lilia1891:
Adjust text for filter label on Indicator dashboard - Asana https://app.asana.com/0/1206017643443542/1208871410132335/f

* Added optional `actionTerm` parameter to `getActionTermContext`;
* Updated `IndicatorList` to pass `actionTerm: 'INDICATOR'` for the correct context;
* Added `INDICATOR` context to the `filter-text` in locale files (EN and DE).

Before:
Screenshot 2025-01-07 at 11 33 06

After:
Screenshot 2025-01-07 at 11 34 41

To do next: Add the rest of translations for `filter-text` via Weblate.

### DIFF
--- a/common/i18n.ts
+++ b/common/i18n.ts
@@ -4,10 +4,15 @@ import {
   SiteGeneralContentActionTerm,
 } from './__generated__/graphql';
 
-export function getActionTermContext(plan: {
-  generalContent?: { actionTerm: SiteGeneralContentActionTerm };
-}) {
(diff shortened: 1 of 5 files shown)


terotik (approved):
Having "indicator" as action term context might become confusing sometime. But I guess it's the best solution nevertheless in this case where we are using the same filters in different context!
Looks good @lilia1891 !

PR #453, opened 2025-01-15
Use graph visual settings from theme

terotik:
Dependency: https://github.com/kausaltech/kausal-themes/pull/39

Before merging Patchenstein, moving some customer specific settings from there to `master`

Only visible change in existing graphs should be `hoverformat`:
BEFORE
Screenshot 2025-01-15 at 19 20 17

AFTER
Screenshot 2025-01-15 at 19 20 34

**Via settings we can now set:**
- Fill for trace markers
- Symbol types for all trace markers
- Colors of all traces
- If goals are connected with line

**User need:** 
- Filled circle markers
- Specific color for main trace and goal trace
- Line connecting goal points

Screenshot 2025-01-15 at 19 23 47

### DIFF
--- a/components/graphs/IndicatorGraph.tsx
+++ b/components/graphs/IndicatorGraph.tsx
@@ -214,7 +214,6 @@ const createTraces: (params: CreateTracesParams) => TracesOutput = (params) => {
   const allXValues = [];
 
   const newTraces = traces.map((trace, idx) => {
-    console.log('trace', trace);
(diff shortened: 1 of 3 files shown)


terotik:
> Very nice improvement! Do you keep console.logs in the code on purpose?

Thanks! No not on purpose. I try to clean them up pre-commit but I always miss some :(

PR #455, opened 2025-01-16
Display search text badge again

lilia1891:
Search text badge is disappeared from the ActionListView filter section - Asana https://app.asana.com/0/1206017643443542/1209138285063386/f
Regression bug: the logic of handling boolean filters could cause `name` filters string values return null.

* Refactored `createBadge` function:
Improved validation for `name` filter, logic for `boolean` and `options` filters

Before:
Screenshot 2025-01-16 at 11 15 21

After:
Screenshot 2025-01-16 at 11 15 50


terotik (approved):
Nice, works smoothly!

PR #459, opened 2025-01-27
Make action links in multiplan redirect to the correct plan

lilia1891:
Multiplan organization page action links are not linked to the correct plans - Asana https://app.asana.com/0/1206017643443542/1209193372014336/f

* Updated `ActionLink` to prepend `planUrl`;
* Added `planViewUrl` as a fallback for cross-plan links.

There is a difference in domain structure between local/testing and production environments for Ludwigsburg (e.g `ludwigsburg-stadtteile` in test and `ziele.ludwigsburg.de/stadtteile` in prod), so I can't test this fix properly.

### DIFF
--- a/common/links.tsx
+++ b/common/links.tsx
@@ -117,7 +117,7 @@ export function ActionLink({
     : action.identifier;
 
   const actionLink = usePrependPlanAndLocale(
-    `${ACTIONS_PATH}/${targetIdentifier}`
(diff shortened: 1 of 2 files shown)


terotik (approved):
Looks good locally. Let's test with this if the fix applies in environments. At least we are not breaking something that isn't already broken.

PR #460, opened 2025-01-28
Apply based on language hyphenation for action.name

lilia1891:
Change `hyphens` from `manual` to `auto` for all languages except Finnish - Asana https://app.asana.com/0/1206017643443542/1208784984047146/f

 We get `hyphens: manual` from backend for `action.name` and need to use that for Finnish language. For other languages should be applied `hyphens: auto`.
 
 * Applied the css `lang(fi)` pseudo-class to the specific components to achieve that.

### DIFF
--- a/components/actions/ActionHighlightCard.tsx
+++ b/components/actions/ActionHighlightCard.tsx
@@ -57,8 +57,12 @@ const StyledCardTitle = styled(CardTitle)`
   font-size: ${(props) => props.theme.fontSizeMd};
   color: ${(props) => props.theme.neutralDark};
   text-align: left;
-  hyphens: manual;
+  hyphens: auto;
   margin-bottom: 0;
+
+  :lang(fi) & {
+    hyphens: manual;
+  }
 `;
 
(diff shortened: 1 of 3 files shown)


components/actions/ActionHighlightCard.tsx
```diff
@@ -57,8 +57,12 @@ const StyledCardTitle = styled(CardTitle)`
   font-size: ${(props) => props.theme.fontSizeMd};
   color: ${(props) => props.theme.neutralDark};
   text-align: left;
-  hyphens: manual;
+  hyphens: auto;
   margin-bottom: 0;
+
+  :lang(fi) & {
```
woodwoerk, line 63: Does this need to have an `&` to apply the `lang` pseudoclass to this element rather than the child elements? (same question for the other uses of this class)

```suggestion
  &:lang(fi) & {
```

woodwoerk: @lilia1891 Thanks for the clarification! In that case, this looks great

woodwoerk (approved):

PR #464, opened 2025-02-03
Remove duplicated types in CategoryTags

lilia1891:
A bug: Category tags are double-counted on Indicator Details Page - Asana https://app.asana.com/0/1206017643443542/1209302839473157
The problem is caused by the types array containing duplicated types

* Removed the unnecessary `types` prop from `CategoryTags`;
* Used `Map` to avoid duplication of category types;

Before:

Screenshot 2025-02-03 at 16 37 45

After:

Screenshot 2025-02-03 at 16 37 21

### DIFF
--- a/components/actions/ActionContent.tsx
+++ b/components/actions/ActionContent.tsx
@@ -289,9 +289,7 @@ function ActionContentBlock(props: ActionContentBlockProps) {
       const categories = action.categories.filter(
         (cat) => cat.type.id === block.categoryType.id
       );
-      return (
-        <CategoryTags categories={categories} types={[block.categoryType]} />
-      );
+      return <CategoryTags categories={categories} />;
     }
     case 'ActionContentAttributeTypeBlock': {
       const attribute = action.attributes.find(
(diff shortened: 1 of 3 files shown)


tituomin (commented):
Thanks! I'm still curious where the duplicates are coming from originally so I'll dig into the code a little bit. I'll be back.

tituomin (changes requested):
I suggest a different approach.

The `types` prop does not actually make any sense.

We are using `CategoryTags` in two places and always either the `categories` prop is derived form the `type` prop, like here:

```
    case 'ActionContentCategoryTypeBlock': {
      const categories = action.categories.filter(
        (cat) => cat.type.id === block.categoryType.id
      );
      return (
        <CategoryTags categories={categories} types={[block.categoryType]} />
      );
    }

```

or vice versa:

```
            <CategoryTags
              categories={indicator.categories}
              types={indicator.categories.map((c) => c.type)}
              noLink={true}
            />
```
I think we should just remove the `type` property and always calculate the unique types inside the CategoryTags almost like you have done (but getting the types first from the categories). In fact, you can just use a `Set` for the types.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

lilia1891:
> I think we should just remove the `type` property and always calculate the unique types inside the CategoryTags almost like you have done (but getting the types first from the categories). In fact, you can just use a `Set` for the types.

@Timo, thanks for the suggestion - I removed the`type`. `Set` didn't work because objects are compared by reference, not value. So I stick with the `Map()`.

PR #465, opened 2025-02-05
Avoid crashing if an image does not have a rendition

bbliem:
An image does not necessarily have a rendition, as per the GraphQL schema, for whatever reason. Our code did not take this into account and would crash if an image has no rendition. This seems to have been the cause for the following errors:

- https://sentry.kausal.tech/organizations/kausal/issues/4039/
- https://sentry.kausal.tech/organizations/kausal/issues/4033/

Corresponding Asana task:
https://app.asana.com/0/1206017643443542/1209261486348088

### DIFF
--- a/common/images.js
+++ b/common/images.js
@@ -22,14 +22,14 @@ export const getBgImageAlignment = (image) => {
 export function getActionImage(plan, action) {
   let image;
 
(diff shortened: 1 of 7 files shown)


Bhavatu (approved):
Makes sense, didn't test though

PR #469, opened 2025-02-21
fix icon badges from clipping text and descending letters

Bhavatu:
Fix width clipping some badges
Also fix descending letters clipping

From this:
image

To this (localhost so no icons showing) :
image


terotik (approved):
Amazing. Thanks @Bhavatu :)

PR #470, opened 2025-02-24
Fix organization term in tooltip

tituomin:
Add more stuff to query.
Also make code not break if query doesn't have enough stuff.


woodwoerk (approved):
Joe like more stuff in query. Joe like code not break.

PR #471, opened 2025-02-25
[1/3] Bundle of code changes to prepare for Paths integration

terotik:
- Enable use of custom components for Header and Footer set from theme
- Show action pagination setting from theme
- Hide action implementation phase completely if not used
- Breadcrumbs: styling and handle multiple main categories
- Accordion: small styling changes
- Add new (Paths) icons in AvailableIcons
- Do not link to indicator list if the indicator list page is not in use
- Simplify indicator graph (draw bar chart only when single year AND multiple categories, handle unit special cases)


woodwoerk:
Code looks good, I just noticed one possible regression. In multiplan setups I'm seeing a 404 for some pages. For example:
- http://longmont-sca.localhost:3000/topic-area/air-quality
- http://sunnydale.localhost:3000/actions/U2

| `master` | `feature/paths-related-changes` |
|--|--|
| image | image |

PR #479, opened 2025-03-18
Feat/primary responsible party filter

lilia1891:
Add an option to filter actions by only the primary responsible party/organization on the `ActionListView` page. 
Asana - https://app.asana.com/0/1206017643443542/1208880507599785/f

* Implemented a toggle to filter actions by the selected organization with the primary role only.
* Made toggle appear only if organization selected.

https://github.com/user-attachments/assets/6d7a613d-63ff-405a-9ace-c29e19ed5e9b

### DIFF
--- a/components/actions/ActionListFilters.tsx
+++ b/components/actions/ActionListFilters.tsx
@@ -668,11 +668,13 @@ class PrimaryResponsiblePartyFilter extends DefaultFilter<string | undefined> {
             checked={value ? true : false}
             onChange={(e) => onChange(this.id, e.target.checked)}
           />
-          <label htmlFor={this.id}>{this.getLabel(t)}</label>
-          <PopoverTip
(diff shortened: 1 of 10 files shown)


terotik:
Looks good! Two notes: @lilia1891
- when the user closes/removes the "as the primary organisation" filter tag - the toggle stays selected. Should unselect the toggle automatically
- Do you think it would be possible to render the toggle directly underneath the dropdown in the same block? I know it would cause the layout to  shift a bit, but would be better for cases where the org dropdown is the last on the line and the toggle appears on next line on the left.

terotik (approved):
Looks good @lilia1891 . I think ActionListFilters file is getting too massive and needs to be split up to smaller components. But that's a housekeeping task for another time. (I'll file a ticket for that). Let's launch this 👍

PR #483, opened 2025-04-04
Fix duplicated category tags on indicator page

lilia1891:
A bug: Category tags are double-counted on Indicator Details Page - https://lappeenranta.test.kausal.tech/lumo/indicators/3894
Asana https://app.asana.com/0/1206017643443542/1209302839473157
The problem is caused by the types array containing duplicate types.

* Added `uniqueTypes` to remove duplicates so only unique types (based on their id) are passed to the CategoryTags.


woodwoerk (approved):
Looks good!

PR #484, opened 2025-04-09
Fix problem with _next/static.* file requests ending up in app code

tituomin:
The way nextjs routing works is that it checks if a static file exists in the filesystem and if it does not, it will move on to routing through the application. This results in missing static files causing errors in app code.

Fix this problem by marking the static file requests specially in the middleware and exiting early from app code if the request is thus marked. We need to modify both the app layout and the middleware, because inside the middleware we do not yet know if the static request will be succesful. Only unsuccesful requests end up in the app code.

### DIFF
--- a/app/[domain]/[lang]/[plan]/layout.tsx
+++ b/app/[domain]/[lang]/[plan]/layout.tsx
@@ -13,6 +13,7 @@ import { SharedIcons } from '@/components/common/Icon';
 import { MatomoAnalytics } from '@/components/MatomoAnalytics';
 import { getMetaTitles } from '@/utils/metadata';
 import { tryRequest } from '@/utils/api.utils';
+import { shouldIgnoreRequest } from '@/utils/middleware.utils';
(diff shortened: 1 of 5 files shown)


app/[domain]/[lang]/[plan]/layout.tsx
```diff
@@ -36,6 +36,16 @@ export async function generateMetadata({ params }: Props): Promise<Metadata> {
...
   }
+
+  const staticFileErrorHeader = headersList.get('x-static-file-error');
+  if (staticFileErrorHeader === 'not-found') {
+    // Due to the internal nextjs implementation, 404 results for
+    // missing files served at _next/static end up here.
+    // We have intercepted all static requests in the middleware and
+    // injected the custom header to be able to exit early here
+    // TODO: remove after having moved to serve the static files outside nextjs
+    notFound();
```
Comment (woodwoerk) [inline app/[domain]/[lang]/[plan]/layout.tsx:47]:
```suggestion
    return {};
```

PR #486, opened 2025-04-10
Fix nextjs static asset routing problems

tituomin:
Add a static path segment to the beginning of app routes so that the dynamic segments of the paths only match actual app requests and nothing else.


woodwoerk (approved):
So simple. So elegant.

PR #490, opened 2025-05-07
Fix years mismatch due to timezone difference

lilia1891:
Indicator years in data table in public UI do not match the years in admin UI  - Asana https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1210078915968237?focus=true

The browser interpreted dates like midnight in Finnish time as the previous day in UTC, which made the displayed year shift back. 

* Added `utc` plugin to `dayjs` setup.
* Used the raw date `rowObj.raw` instead of formatted label for trace lookups.


tituomin:
The way we handle dates coming from the backend is very convoluted (even before reaching this GraphAsTable component), so I think I need to take another look and see if it could be cleaned up.

PR #493, opened 2025-05-14
Handle reverse progression on indicator showcase

terotik:
Handle an edge case when emissions have increased from the base value. Normalize by default in this case.
There is still a glitch in animation when toggling normalisation on/off.

Before:
Screenshot 2025-05-14 at 20 41 27

After:
Screenshot 2025-05-14 at 20 41 18


woodwoerk (approved):
Nice! Looks good

PR #495, opened 2025-05-15
Add initial boilerplate for indicator dashboard blocks

woodwoerk:
This is mainly boilerplate so that we can work on this simultaneously. Card styles and such will need to be worked on.

- Add initial query to fetch the new indicator dashboard blocks
- Update the streamfield component to render the dashboard row and child blocks

→ Check the admin content which renders the below view [here](https://admin.watch.kausal.tech/admin/pages/6431/edit/)
→ Test at http://sunnydale.localhost:3000/about-us-temp

Screenshot 2025-05-15 at 18 36 31


terotik (approved):
Looking good! Great starting point 💪

PR #496, opened 2025-05-16
Update checkbox to be unchecked by default

lilia1891:
Longmont request - `Let’s have the default setting be the opposite, so the “don’t show again” is unclicked. I don’t know how frequently visitors visit the site, so I want it to pop-up unless they ask it not to.`


terotik (approved):
👍👍👍

PR #497, opened 2025-05-16
Fix/remove dependencies from mass query

tituomin:
Previously, the `ActionCardList` has queried all of the dependency related fields of all of the actions in the mass query `GET_ACTION_LIST`. It is however unecessary, and it probably introduces unnecessary complexity to optimize this on the backend side. So we added to the backend a boolean `hasDependencyRelationships` which is optimized for mass queries.

In this PR, the UI uses the `hasDependencyRelationships` field to see if the dependency icon should be displayed and only when the popup for the dependencies is rendered do we actually start querying for the dependency related fields for the action.

We replace enriching all of the actions with the necessary fields in a loop with a method that can be used to query the full details of any action in the context of the popup (where we only know the action's dependencyRole and it's relationship to the current action).


components/actions/ActionCardList.js
```diff
@@ -118,6 +118,12 @@ function ActionCardList({
     theme,
     t
   );
+  const actionsById = useMemo(
+    () => new Map(actions.map((a) => [a.id, a])),
+    [actions]
```
woodwoerk, line 123: It looks like `actions` is a new array on each render, so this memoization isn't necessary. If we notice performance issues we should refactor `ActionList`, it looks pretty messy anyway 👀

components/actions/blocks/action-dependencies/ActionDependenciesBlock.tsx
```diff
@@ -22,16 +27,45 @@ type ActionGroup = {
...
 type Props = {
+  action: Action;
   activeActionId?: string;
   size?: 'default' | 'small';
-  actionGroups: ActionGroup[];
   showTitle?: boolean;
   title?: string;
   helpText?: string;
+  getFullAction: (id: string) => Action;
+  skeleton?: boolean;
```
woodwoerk, line 37: Nitpick: Consider renaming `skeleton` to `loading` to reflect the state rather than its visual representation

components/actions/blocks/action-dependencies/ActionDependenciesBlock.tsx
```diff
@@ -106,8 +141,15 @@ export function mapActionToDependencyGroups(
     []
   );
   const uniqueActions = uniqBy(flatDependencyList, 'id');
+  let expandedActions = uniqueActions;
+  if (getFullAction != null) {
+    expandedActions = uniqueActions.map((a) =>
+      Object.assign({}, getFullAction(a.id), a)
+    );
+  }
```
woodwoerk, line 149: Nitpick: preferring `const` over `let` is accepted as best practice in JS lore, instead we could do:

```suggestion
  const expandedActions = getFullAction
    ? uniqueActions.map((a) => Object.assign({}, getFullAction(a.id), a))
    : uniqueActions;
```

components/actions/blocks/action-dependencies/ActionDependenciesBlock.tsx
```diff
@@ -122,21 +164,105 @@ export function mapActionToDependencyGroups(
...
+    skeleton === true ||
+    (action?.dependencyRole != null &&
+      action?.allDependencyRelationships != null);
+
+  const { workflow } = useWorkflowSelector();
+
+  const { error, data } = useSuspenseQuery<ActionDependenciesQuery>(
+    GET_ACTION_DEPS,
+    skipFetchingDependencies
+      ? skipToken
```
woodwoerk, line 215: Nice! TIL

woodwoerk (commented):
Nice work! Code looks great, nice use of suspense. Could you deploy the backend so that I can test the UI locally before approving?

Dropped a few minor nitpicks, nothing that would stop merging this.

PR #499, opened 2025-05-19
Add indicator pie chart

woodwoerk:
Add the indicator pie chart component and update the dashboard row styles. Parts of this chart will may be useful for the other visualisations. Let's separate those to helper functions/ constants as needed.

### Pie chart with colours defined in dimension categories
| Input | Output |
|--|--|
| image |  image |
| image | image) |

### Pie chart with no colours defined in dimension categories (defaults to theme)
| Input | Output |
|--|--|
|  image | image |

### Invalid pie input

| Input | Output |
|--|--|
| image | image |

### DIFF
--- a/common/__generated__/graphql.ts
+++ b/common/__generated__/graphql.ts
@@ -195,6 +195,7 @@ export type Action = {
   editUrl?: Maybe<Scalars['String']['output']>;
   /** The date when implementation of this action ends */
   endDate?: Maybe<Scalars['Date']['output']>;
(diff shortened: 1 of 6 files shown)


terotik (approved):
Nice! Excited about this feature 🤩
I was wondering should we test out [echarts aria option](https://echarts.apache.org/handbook/en/best-practices/aria/) for accessibility? Would be much better than tabular data in the dashboard context.

PR #504, opened 2025-05-30
Feat/indicator line chart

lilia1891:
Implement `DashboardIndicatorLineChartBlock` component that renders line charts for different indicator types using ECharts. 
 Asana - https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1210259161444274?focus=true.
 
 
one-parameter
one-dim-with-cats
multi-dims-with-shared-cats

### DIFF
--- a/components/contentblocks/indicator-chart/DashboardIndicatorLineChartBlock.tsx
+++ b/components/contentblocks/indicator-chart/DashboardIndicatorLineChartBlock.tsx
@@ -2,6 +2,7 @@
 
 import React from 'react';
 import { useTheme } from 'styled-components';
+import { useTranslations } from 'next-intl';
 import { LineChart, ScatterChart } from 'echarts/charts';
 import {
   GridComponent,
@@ -39,19 +40,20 @@ const X_SYMBOL =
   '-0.192399 20.7783 0.979174L13.0001 8.75735L5.22191 0.979175C4.05033 -0.192398 2.15084 -0.192398 0.979266 0.979175C-0.192307 ' +
   '2.15075 -0.192307 4.05024 0.979266 5.22182L8.75744 13L0.979266 20.7782Z';
 
-const DashboardIndicatorLineChartBlock: React.FC<Props> = ({
+const DashboardIndicatorLineChartBlock = ({
   chartSeries,
   indicator,
   dimension,
   showTotalLine,
-}) => {
+}: Props) => {
   const theme = useTheme();
+  const t = useTranslations();
   const graphsTheme: GraphsTheme = theme.settings?.graphs ?? {};
   const unit = indicator?.unit?.name ?? '';
   const palette = graphsTheme.categoryColors ?? getDefaultColors(theme);
 
   if (!chartSeries?.length) {
-    return <div>No data</div>;
+    return <div>{t('data-not-available')}</div>;
   }
 
   const dimCategoryMap = new Map<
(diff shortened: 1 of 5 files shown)


components/contentblocks/indicator-chart/DashboardIndicatorLineChartBlock.tsx
```diff
@@ -0,0 +1,263 @@
...
+        const rows = ary
+          .filter((p) => {
+            if (!legendData.includes(p.seriesName)) return false;
+            if (seen.has(p.seriesName)) return false;
+            seen.add(p.seriesName);
+            return true;
+          })
+          .map((p) => {
+            const v = (p.data as any[])[1] as number;
+            return `${p.marker} ${p.seriesName}: ${v.toFixed(1)} ${unit}`;
```
woodwoerk, line 233: `XX.0` on all values looks a bit strange, and some values might be more precise fractions where 1dp is not enough detail. Ideally the precision would be defined on the backend and we'd round the number to that, but I don't think that's currently supported. I suggest we just don't round these numbers for now and show them exactly as precise as the admins entered, what do you think?

image

woodwoerk: Also, ideally the tooltip marker would be the same shape as on the graph. I remember that being tricky in ECharts though, so let's leave it as-is if it's too much of a pain

image

components/contentblocks/indicator-chart/DashboardIndicatorLineChartBlock.tsx
```diff
@@ -0,0 +1,263 @@
...
+      bottom: 0,
+      textStyle: { color: theme.textColor.secondary },
+    },
+    tooltip: {
+      trigger: 'axis',
+      appendTo: 'body',
+      axisPointer: { type: 'line' },
+      formatter: (params) => {
+        const seen = new Set<string>();
+        const ary = Array.isArray(params) ? params : [params];
```
woodwoerk, line 222: Nitpick: I found this a bit hard to read at first because of the naming, suggestion: 
```suggestion
        const processedSeries = new Set<string>();
        const paramsArray = Array.isArray(params) ? params : [params];
```

components/contentblocks/indicator-chart/DashboardIndicatorLineChartBlock.tsx
```diff
@@ -0,0 +1,263 @@
...
+  dimension,
+  showTotalLine,
+}) => {
+  const theme = useTheme();
+  const graphsTheme: GraphsTheme = theme.settings?.graphs ?? {};
+  const unit = indicator?.unit?.name ?? '';
+  const palette = graphsTheme.categoryColors ?? getDefaultColors(theme);
+
+  if (!chartSeries?.length) {
+    return <div>No data</div>;
```
woodwoerk, line 54: Please stick this in the translations

components/contentblocks/indicator-chart/DashboardIndicatorLineChartBlock.tsx
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  const dimCategoryMap = new Map<
```
woodwoerk, line 57: There's a lot going on in this component, I think we could split it up into smaller functions (`getDimSeries`, `getTotalSeries` etc). But we can leave that for V2 and the other graphs.

PR #505, opened 2025-06-06
Style dashboard cards and limit hover effect to clickable cards

terotik:
- Make dashboard cards stand out with a row background color
- Limit card hover effect only to linked cards


lilia1891 (approved):
Nice improvement!

PR #506, opened 2025-06-06
Fix: remove global height overriding iframe height

lilia1891:
Bug fix - Embeds were displayed at the wrong height.
Asana https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1210470076974007?focus=true

Remove the global height: 100% style for iframes to fix overriding issues with embeds


terotik (approved):
Easy fix :)

PR #508, opened 2025-06-11
Bar and area charts for indicator visualisation 

lilia1891:
Add the new blocks for the bar and area charts based on e-charts for indicators visualisation. 

Asana tasks: https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1210259161444271?focus=true and https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1210259161444275?focus=true.

* Implemented Bar Chart Block:
  - Dynamically renders grouped or stacked bars based on the admin settings;
  - Supports both total and dimensions with categories.
  - Bar charts don't support trendlines or goal markers.
 * Implemented Area Chart Block:
     - Supports both total and dimensional indicators.
     - Displays total and trendline only when no dimension is selected (for total values only).
     - Don't support goal markers.
  * Added support for the admin indicator visualisation settings: `minValue`, `maxValue`, `valueRounding`, `ticksCount`, `ticksRounding`.
  
  Bar charts (grouped):
  
bar-charts-grouped

Bar charts (stacked):

bar-charts-stacked

Area charts:

area-charts

### DIFF
--- a/components/contentblocks/DashboardRowBlock.tsx
+++ b/components/contentblocks/DashboardRowBlock.tsx
@@ -15,6 +15,7 @@ import DashboardIndicatorSummaryBlock from './DashboardIndicatorSummaryBlock';
 import DashboardIndicatorPieChartBlockComponent from './indicator-chart/DashboardIndicatorPieChartBlock';
 import DashboardIndicatorLineChartBlockComponent from './indicator-chart/DashboardIndicatorLineChartBlock';
 import DashboardIndicatorBarChartBlockComponent from './indicator-chart/DashboardIndicatorBarChartBlock';
+import DashboardIndicatorAreaChartBlockComponent from './indicator-chart/DashboardIndicatorAreaChartBlock';
 
 const DashboardRowSection = styled.div<{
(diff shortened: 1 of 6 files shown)


woodwoerk:
These look SO GOOD! 🔥 I've only just started testing, but I noticed one thing so far, I'll start looking through the code too.

There seem to be some issues with the values in the area chart. I'm not sure if it's the tooltip or chart showing incorrect values, but in the below example you can see "Good" is 1000km for 2023 but the chart point is around 970.

image

lilia1891:
Good catch! That happened because in the admin vis settings I set some values (`Tick rounding` and `Value rounding`).

woodwoerk (approved):
Great work, I think these are looking really good! Nice splitting up of the utility functions.

I think it's fine to leave goals and trend lines out of these charts for now. It makes sense to add them but we can do that in another task. And for the other properties we discussed, "desired trend" isn't necessary and the stacking works well with the `barType === 'stacked'` check so we can ignore the other stackable property.

PR #512, opened 2025-06-26
Make gridLeft dynamic based on unit name length

lilia1891:
Unit label on Y-axis cropped if it's long  - Asana https://app.asana.com/1/1201243246741462/project/1209894862159781/task/1210615499515662?focus=true

* Applied padding to the y-axis label to prevent cropping

Before:
Screenshot 2025-06-26 at 15 17 23

After:
Screenshot 2025-06-26 at 15 15 50

### DIFF
--- a/components/contentblocks/indicator-chart/DashboardIndicatorAreaChartBlock.tsx
+++ b/components/contentblocks/indicator-chart/DashboardIndicatorAreaChartBlock.tsx
@@ -120,7 +120,13 @@ const DashboardIndicatorAreaChartBlock = ({
         indicator?.valueRounding
       ),
     },
-    grid: { left: 20, right: 20, top: 40, bottom: 60, containLabel: true },
+    grid: {
+      left: 20,
+      right: 20,
+      top: 40,
+      bottom: 60,
+      containLabel: true,
+    },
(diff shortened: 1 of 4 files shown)


terotik:
Instead of adjusting the margin, could we try just adjusting the placement of y-axis name?
Something like this could work:

```   
yAxis: {
  nameTextStyle: {
        align: "left",
        padding: [0, 0, 0, -25],
  }
}
```

Left-align aligns it on the right side of the axis line, and negative padding brings it back a little bit to "fake-align" with the axis numbers. Of course we can not be exact as tick numbers can be different lengths, but this could be close enough?

terotik (changes requested):
Instead of adjusting the margin, could we try just adjusting the placement of y-axis name?
Something like this could work:

```   
yAxis: {
  nameTextStyle: {
        align: "left",
        padding: [0, 0, 0, -25],
  }
}
```

Left-align aligns it on the right side of the axis line, and negative padding brings it back a little bit to "fake-align" with the axis numbers. Of course we can not be exact as tick numbers can be different lengths, but this could be close enough?

lilia1891:
@terotik done, thanks!

PR #514, opened 2025-07-01
Add missing translations for the data-not-available string

lilia1891:
Add "data-not-available" locale string used in indicator graphs for all languages.


woodwoerk (approved):
I can't approve the translations themselves :D but the code looks good. @terotik how's the Finnish translation?

PR #519, opened 2025-07-16
Use new indicator axis settings on indicator graph

terotik:
Use the  new 
```
ticksCount
ticksRounding
valueRounding
```
settings in indicator visualisation to improve y axis rendering
(less duplicate values, user control on rounding)

### DIFF
--- a/components/graphs/IndicatorGraph.tsx
+++ b/components/graphs/IndicatorGraph.tsx
@@ -5,7 +5,7 @@ import React from 'react';
 import { splitLines } from 'common/utils';
 import { merge } from 'lodash';
 import dynamic from 'next/dynamic';
-import { Data, Layout } from 'plotly.js';
+import type { Data, Layout, PlotData } from 'plotly.js';
 import { transparentize } from 'polished';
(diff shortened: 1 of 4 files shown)


components/graphs/IndicatorGraph.tsx
```diff
@@ -40,9 +40,12 @@ const createLayout = (
     yaxis: {
       automargin: true,
       hoverformat: ',.3r',
-      tickformat: ',.1r',
+      tickformat: yRange.ticksRounding
```
woodwoerk, line 43: Could `ticksRounding` be `0`? And if so, should this be:

```suggestion
      tickformat: typeof yRange.ticksRounding === 'number'
```

woodwoerk (approved):
Code looks good, please just check that one comment. I haven't tested this locally.

components/graphs/IndicatorGraph.tsx
terotik: Brilliant spot, thanks @woodwoerk !

PR #526, opened 2025-08-20
Feat/indicator pp tooltip

lilia1891:
Change label 'Time to goal' in indicator summary to 'To Goal' and add a tooltip to explain what a percentage point + link to Wikipedia.
Asana - https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1210886272100508?focus=true

* Changed the IndicatorSummary label "Time to goal" → "To goal",
* Added a tooltip (re-used PopoverTip) next to the p.p. unit when indicators are measured in percentage points.
* Tooltip explains what a percentage point means plus contents link to Wiki (Read more).
* Added translations.

Screenshot 2025-08-20 at 16 29 02


terotik (approved):
I had doubts about this solution, but I think it makes sense after all. Looking good thanks!

PR #528, opened 2025-09-03
Fix linking of related affect/effect indicators when indicator is not part of current plan

Bhavatu:
https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1210812059821647

### DIFF
--- a/src/components/indicators/IndicatorCard.tsx
+++ b/src/components/indicators/IndicatorCard.tsx
@@ -171,20 +171,19 @@ function IndicatorValue({
   );
 }
 
-function CardLink(props) {
+interface CardLinkProps {
+  level?: string | null;
+  indicatorId: string;
+  children: React.ReactNode;
+  customHref: string | null;
+}
+
+function CardLink(props: CardLinkProps) {
   const { level, indicatorId, children, customHref } = props;
 
   if (level !== 'action') {
-    if (customHref) {
-      return (
-        <IndicatorLink id={indicatorId} viewUrl={customHref}>
-          <StyledLink href>{children}</StyledLink>
-        </IndicatorLink>
-      );
-    }
-
     return (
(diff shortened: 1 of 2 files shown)


src/components/indicators/IndicatorCard.tsx
```diff
@@ -172,14 +172,23 @@ function IndicatorValue({
 }
 
 function CardLink(props) {
-  const { level, indicatorId, children } = props;
+  const { level, indicatorId, children, customHref } = props;
```
woodwoerk, line 175: Could you add the types for these?

src/components/indicators/IndicatorCard.tsx
```diff
@@ -200,6 +209,7 @@ interface IndicatorCardProps {
     unit: string;
   } | null;
   disabled?: boolean | null;
+  customHref?: string | null;
```
woodwoerk, line 212: Nitpick: We only need to type for `null` _and_ `undefined` when the value comes from the backend and could be either. Since we create this type we can settle on one or the other.

We use `null` to represent the intentional absence of a value and  `undefined` when a variable hasn't been assigned a value. 

```suggestion
  customHref: string | null;
```

woodwoerk: Also the default parameter `= null` isn't necessary since the value is always either a string or null.

woodwoerk (approved):
Nice! I added a few comments, please at least address the non-nitpicky ones and then feel free to merge. Oddly I wasn't able to check out the branch 🤔 so couldn't test locally.

```
joesmallwood kausal-watch-ui % git fetch
joesmallwood kausal-watch-ui % git checkout fix/multiplan_related_indicators_linking
error: pathspec 'fix/multiplan_related_indicators_linking' did not match any file(s) known to git
```

PR #529, opened 2025-09-04
Override legends for area charts to match pie

lilia1891:
Indicators Dashboard: Visually align area chart legends with pie (or bar) charts.
Asana - https://app.asana.com/1/1201243246741462/project/1209894862159781/task/1210912656811919?focus=true

* Override area charts legends to use `roundRect` icons (like pie-charts).

Before:

before

After:

after


woodwoerk (approved):
Looks good!

PR #707, opened 2026-04-07
Add pledge cards to action detail page

bbliem:
Display related pledges on action pages via the new ActionPledgesBlock content block, with full commit/uncommit interaction support.

## Related issue

[Asana task](https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1213893804379191?focus=true)

[Backend PR that defines the required block](https://github.com/kausaltech/kausal-watch/pull/550)

---

### Type of Change

- [ ] Set the PR's label to match the nature of this change

### Testing

### Dependencies


woodwoerk (changes requested):
Sorry there was a link missing to the designs here, but we should exclude the ability to commit to a pledge from this card. Otherwise this looks good, the commit button should just be hidden.

If you're interested in why we prefer that the user navigates to the pledge first and then commits, this is because:
- Users of the full action plan and pledges belong to different user groups
- We want users to be fully aware of the pledge concept before committing, so a bit of friction to visit the pledge first is useful
- We want to keep interactive pledge elements scoped to the pledge list page and individual pledge pages for clearer hierarchy with our upcoming pledge user auth

PR #717, opened 2026-04-15
Fix/attribute chip contrast

terotik:
- `brandLight` can possibly be dark so we must choose text color for action attribute chip via `readableColor` check
- Make stories for different variations of action attributes
- Up themes


woodwoerk (approved):
I forget that `brandLight` can actually be dark 😄 that gives me the heebiejeebies a little

PR #718, opened 2026-04-16
Fix pledge search bar height when there are no other filters

woodwoerk:
## Description

When there are no additional filters, the search bar didn't stretch to the full height of the container.

| Before | After |
|-|-|
| image | image |


lilia1891 (approved):
Niceeee

PR #727, opened 2026-04-28
Encode actions variable in action report API call

MWargelin:
## Description
This is an attempt to fix a bug where sometimes the actions variable has a single quote character (') appended to it, making the call crash in the backend.

[Here is an example of the issue in Sentry](https://sentry.kausal.tech/organizations/kausal/issues/10732/?project=3&query=assigned%3Ame&referrer=issue-stream), where the API call has ended up in backend and the last action id has the single quote character appended, and it crashes when the id is processed as integer.

This was a Claude-generated answer. It thought for a long time (~10 minutes) and got no other ideas. I'm guessing if it's not this, then somehow backend might return the ids wrong, but I couldn't figure out how that could happen.

I couldn't reproduce the bug, so this is a somewhat blind guess. If you think it doesn't make sense, please tell, I'm willing to believe that! :smile: 

\-

## Related issue
[Sentry](https://sentry.kausal.tech/organizations/kausal/issues/10732/?project=3&query=assigned%3Ame&referrer=issue-stream)

\-

## Additional Notes
\-

---

### Testing

  \-

- [x] **New strings are translatable** (all user-facing text uses i18n)

### Dependencies

### DIFF
--- a/src/components/dashboard/ActionStatusExport.js
+++ b/src/components/dashboard/ActionStatusExport.js
@@ -8,8 +8,8 @@ export default function ActionStatusExport({ actions }) {
   const plan = usePlan();
   const url = plan.actionReportExportViewUrl;
   const actionIds = actions.map(({ id }) => id).join(',');
-  const csvExportUrl = `${url}?actions=${actionIds}&format=csv`;
-  const excelExportUrl = `${url}?actions=${actionIds}&format=xlsx`;
+  const csvExportUrl = `${url}?actions=${encodeURIComponent(actionIds)}&format=csv`;
+  const excelExportUrl = `${url}?actions=${encodeURIComponent(actionIds)}&format=xlsx`;
   return (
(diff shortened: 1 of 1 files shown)


tituomin:
I don't think the encoding helps. In the situation where that quote char has ended up in the URL, I cannot see a way where the quote char already wasn't part of the actionIds string, the last character of that string. Encoding the string only encodes the commas inside that string (which is not required) but does not help with the quote. It will look like this: "9738%2C9739%2C9740%2C9741%2C9742%2C10165%2C10664%2C15834%2C10207%2C10208%2C10952%2C10210%2C10211%2C10215%2C10230%2C10231%2C10232%2C10233%2C10234%2C9179%2C10238%2C10239%2C10251%2C10252%2C14822%2C10253%2C10254%2C10255%2C10256%2C10257%2C10258'"

PR #730, opened 2026-04-29
De-index non-production deployments via X-Robots-Tag header

Bhavatu:
## Description
The existing `Disallow: /` in robots.ts on non-production deployments only blocks crawling, not indexing. URLs can still appear in search results via inbound links, with no description. Switch to emitting `X-Robots-Tag: noindex, nofollow` from middleware so crawlers actually drop the URLs from their indexes once they re-crawl.

For the directive to be seen, crawlers must be allowed to fetch the page, so robots.ts now allows `/` on every deployment. The sitemap is gated to production so non-production stops advertising its URLs.

Seems counterintuitive, but the i.e. google [docs](https://developers.google.com/search/docs/crawling-indexing/block-indexing) say:
>Important: For the noindex rule to be effective, the page or resource must not be blocked by a robots.txt file, and it has to be otherwise accessible to the crawler. If the page is blocked by a robots.txt file or the crawler can't access the page, the crawler will never see the noindex rule, and the page can still appear in search results, for example if other pages link to it.

Add screenshots or videos demonstrating the changes if applicable.

## Related issue
https://kausaltech.slack.com/archives/C05B87BS9NY/p1777458650827369

Describe or link possible requirements, dependencies and related PRs here

## Additional Notes

Any additional information that reviewers should know about this PR.

---

### Type of Change

- [ ] Set the PR's label to match the nature of this change

### Testing

### Dependencies

### DIFF
--- a/src/app/root/[domain]/[lang]/layout.tsx
+++ b/src/app/root/[domain]/[lang]/layout.tsx
@@ -4,6 +4,7 @@ import type { ReactNode } from 'react';
 import { headers } from 'next/headers';
 import Script from 'next/script';
 
+import type { Metadata } from 'next';
 import { NextIntlClientProvider, useMessages } from 'next-intl';
 import 'react-medium-image-zoom/dist/styles.css';
 
@@ -12,6 +13,7 @@ import { EmotionRegistry } from '@common/themes/StyledComponentsRegistry';
 import '@common/themes/styles/main.scss';
 
 import { DayjsLocaleProvider } from '@/common/dayjs';
+import { deploymentType } from '@/common/environment';
 import { ApolloWrapper } from '@/components/providers/ApolloWrapper';
 import { AuthProvider } from '@/components/providers/AuthProvider';
 import { auth } from '@/config/auth';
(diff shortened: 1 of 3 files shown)


woodwoerk (commented):
Can you try using Next's built in `metadata` for this? You can export the metadata at a top level layout component to have finer control over the robots config. As long as we have access to the `deploymentType` there I think it'll work nicely. Updating it in the middleware feels a bit hacky.

https://nextjs.org/docs/app/api-reference/functions/generate-metadata#robots

src/app/root/[domain]/[lang]/layout.tsx
```diff
@@ -21,6 +23,13 @@ type Props = {
   children: ReactNode;
 };
 
+// Non-production deployments tell crawlers not to index any page.
+// `robots.ts` allows crawl on every deployment so the directive is actually seen;
+// child pages/layouts can override `robots` to opt back in if ever needed.
+export const metadata: Metadata = {
+  robots: deploymentType === 'production' ? undefined : { index: false, follow: false },
```
Comment (tituomin) [inline src/app/root/[domain]/[lang]/layout.tsx:30]: This hasn't been yet revealed anywhere so you couldn't know, but we will very soon be moving away from doing checks like "deploymentType === 'production'" to distinguish between production (published) and preview (onboarding) sites. The reason is that the UI deployment for both will actually be the same.

In the future, it will be best to control behavior like this based on explicit fields in the Plan or PlanDomain objects in the GraphQL query. (That would also make it more flexible, say if we wanted to control indexing with a dynamic setting in the backend)

However, at this point we can also go with this approach and just make a note to remember to revisit the indexing when the refactoring that removes deplomentType checks is ready. So maybe just FYI.

tituomin (approved):
I left a note just mabe to remember to do a followup for this later. But I think this is fine for now, to get it done quickly!

PR #734, opened 2026-05-05
Chore/big common upgrade

terotik:
## Description

Use updated [kausal common](https://github.com/kausaltech/kausal-ui-common/pull/19)

- Use Next 16 and migrate to common nextjs config
- Apollo 3 -> 4, changes a lot of import
- Typescript issues baseline á la Paths

[kausal common UI](https://github.com/kausaltech/kausal-ui-common/pull/19)


configs/tsconfig.next.json
```diff
@@ -0,0 +1,30 @@
...
+      }
+    ],
+    "paths": {
+      "@/public/*": ["../public/*"],
+      "@/*": ["../src/*"],
+      "@common/*": ["../kausal_common/src/*"]
+    },
+    "checkJs": true,
+    "allowImportingTsExtensions": true,
+    // TODO: enable full `strict` (drop these overrides) — see paths-ui's tsc-baseline workflow
```
woodwoerk, line 18: @terotik I thought JSON files don't support comments, is this working as expected?

terotik: Apparently 
"tsconfig.json files don't use standard JSON. TypeScript parses them as JSONC (JSON with Comments), which allows // and /* */ comments and trailing commas. This is documented behavior in the TypeScript compiler."

I wanted the comments there for future reference. We want eventually paths-ui strictness to watch-ui, but did not want to include the rework on this pr

PR #736, opened 2026-05-06
Bump dependencies, fix lint

terotik:
- `pnpm lint` — runs full ESLint, surfaces all 4775 issues (used while fixing things).
- `pnpm lint:baseline` — uses `eslint-baseline.json` (68 KB / 2835 lines) to suppress pre-existing errors. Currently passes (0 errors, 256 deprecation warnings — warnings don't fail). This is the script CI should run.
- `src/stories/tsconfig.json` (new) — mirrors paths-ui's stories/tsconfig.json so typescript-eslint's project service can parse story files (they're excluded from tsconfig.next.json).
- `eslint-baseline.json` is now in the working tree — commit it so CI can use it.


woodwoerk (approved):
Tested locally and looks good, perhaps just add `kausal_common/**` to the ts and eslint ignore files

PR #739, opened 2026-05-08
Fix secondary nav clipping when there is no body content

woodwoerk:
## Description

When there's no body content, the `absolute` secondary navigation takes no vertical space on the page and gets cut off by the footer. In these cases, render the secondary nav as relative instead.

### Before
image

### After 
image

## Related issue

https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1213990915317190?focus=true


lilia1891 (approved):
tested a few plans apart from Minneapolis and all good! thank you!

PR #752, opened 2026-05-21
Fix/action tasks block heading

lilia1891:
## Description

Support for using a custom title and help text for the ActionTasksBlock on action details pages.
In admin we provide the option to configure custom fields (heading and help text to show instead of the default), but we didn't have support for them on frontend side and always displayed Tasks as header.

Screenshot 2026-05-21 at 12 55 00

* ActionTasksBlock now uses the configured custom title and help text when available;
* Keep the existing “Tasks” label as a fallback.

Screenshot 2026-05-21 at 12 57 17

## Related issue

[slack](https://kausaltech.slack.com/archives/C029WNCMF0T/p1779280873568449)

none

## Additional Notes

Any additional information that reviewers should know about this PR.

---

### Type of Change

- [fix] Set the PR's label to match the nature of this change

### Testing

 Can be tested on action details page of beaverton-climate plan.

### Dependencies


src/components/actions/blocks/ActionTasksBlock.tsx
```diff
@@ -6,16 +6,26 @@ import { ActionSection, SectionHeader } from '@/components/actions/ActionContent
 import TaskList from '@/components/actions/TaskList';
 import { usePlan } from '@/context/plan';
 
-const ActionTasksBlock = (props) => {
-  const { tasks } = props;
+import PopoverTip from '@/components/common/PopoverTip';
+
+type ActionTasksBlockProps = {
+  tasks: React.ComponentProps<typeof TaskList>['tasks'];
```
woodwoerk: Does this work even though `TaskList` is a JS file with no type definitions? It seems based on the legacy [`propTypes` definition that ](https://github.com/kausaltech/kausal-watch-ui/blob/8954bf92f163764108fcfdd4b9877be08c4e4069/src/components/actions/TaskList.js#L229-L231) that `tasks` is just an array of anything anyway, so perhaps we could make this stricter here.

woodwoerk (approved):

src/components/actions/blocks/ActionTasksBlock.tsx
lilia1891: It works, but you're absolutely right. I updated it to use the generated GraphQL type from GetActionDetailsQuery['action']['tasks'] instead. Thank you!

PR #773, opened 2026-06-24
Use plan.name as fallback if plan.shortName is missing

tituomin:
Applies to merged actions if they are from another plan and that plan has no short name.

## Description

Use plan.name as fallback if plan.shortName is missing

Before:

image

After:
"null" is replaced with plan name.

## Related issue

https://app.asana.com/1/1201243246741462/project/1210042120863961/task/1215962515384079?focus=true

-

## Additional Notes

The viewUrl is there to fix a pre-existing type error which now surfaced from baseline due to suffling some types/fields around.

---

### Type of Change

- [X] Set the PR's label to match the nature of this change

### Testing

- [X] **Unit tested**
- [X] **Manually tested locally** (functionality verified)

- [X] **New strings are translatable** (all user-facing text uses i18n)
- [X] **Accessibility** standards met (WCAG compliance, screen reader support)

### Dependencies

- [X] **Dependencies are merged** (if applicable. If the change depends on other PRs e.g. kausal_common)

### DIFF
--- a/src/utils/action.utils.ts
+++ b/src/utils/action.utils.ts
@@ -1,0 +1,18 @@
+interface MergedWithLike {
+  identifier: string;
+  plan: {
+    id: string;
+    shortName?: string | null;
+    name?: string | null;
+  };
+}
+
+export function getMergedName(mergedWith: MergedWithLike, planId: string): string {
+  if (mergedWith.plan.id === planId) {
+    return mergedWith.identifier;
+  }
+
+  const planLabel = mergedWith.plan.shortName || mergedWith.plan.name;
+
+  return planLabel ? `${planLabel} ${mergedWith.identifier}` : mergedWith.identifier;
+}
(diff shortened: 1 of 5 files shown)


src/utils/action.utils.ts
```diff
@@ -0,0 +1,18 @@
...
+    id: string;
+    shortName?: string | null;
+    name?: string | null;
+  };
+}
+
+export function getMergedName(mergedWith: MergedWithLike, planId: string): string {
+  if (mergedWith.plan.id === planId) {
+    return mergedWith.identifier;
+  }
```
woodwoerk, line 13: What's the rationale behind returning an identifier here rather than a name? I'd expect `getMergedName` to always return the name of a plan.

woodwoerk: Actually, I see that this follows the original logic so you might not be able to answer that

woodwoerk (approved):
`null`

src/utils/action.utils.ts
tituomin: It's returning a "display name" for an action, not plan. It's because when an action is merged with an action from the same plan we are looking at, we are assuming it's enough to identify the action with just the id, otherwise we need the plan name as a "namespace"

PR #775, opened 2026-06-26
Fix umbrella action category filtering

bbliem:
Umbrella action lists need to compare categories through common category IDs, but the frontend only enabled common-category remapping when the primary action classification had a common category type. Plans such as lpr-kestavyys-copy1 use a primary classification without a common mapping, so child-plan local category IDs were compared against common filter option IDs and no actions matched. Pass related-plan mode explicitly into action category mapping, match common primary category types by identifier, and add a regression test for remapping child-plan categories through their common category.

## Related issue

[Slack thread](https://kausaltech.slack.com/archives/C031Z5WMGJ0/p1782459886043429)

---

### Type of Change

- [x] Set the PR's label to match the nature of this change

### Testing

- [x] **Built E2E tests** (if applicable. E2E tests added/updated)

- [x] **New strings are translatable** (all user-facing text uses i18n)

### Dependencies

### DIFF
--- a/src/common/__tests__/categories.test.ts
+++ b/src/common/__tests__/categories.test.ts
@@ -1,0 +1,87 @@
+import { constructCatHierarchy, mapActionCategories } from '@/common/categories';
+
+describe('constructCatHierarchy', () => {
+  it('preserves parent paths when mapping categories to common categories', () => {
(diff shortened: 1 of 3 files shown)


woodwoerk (approved):
Looks good!

PR #789, opened 2026-07-16
Fix hydration mismatches from unpinned time rendering

bbliem:
Hydration errors were firing across many tenants and page types (WATCH-UI-3Q6), because time was rendered during SSR in ways the client couldn't reproduce:

- Relative time (`dayjs(...).fromNow()`) depends on the current time, so the value computed at SSR could differ from the one at hydration. This drove the bulk of the occurrences via the actions-list "updated at" cell.
- No global time zone was configured, so next-intl and dayjs formatted timestamps in the server's zone during SSR and the browser's zone on the client, mismatching for users outside the server's zone.

Changes:
- Add a `TimeZoneProvider` client wrapper that pins next-intl's `timeZone` to the plan's zone. It renders the client variant of NextIntlClientProvider so it inherits messages/formats from the parent instead of re-serializing them.
- Format the ChangeHistory timestamp in the plan's time zone (it's an absolute instant, so `.tz()` is deterministic across the server and client).
- Load the dayjs `timezone` plugin.
- Suppress the hydration warning on the relative-time UpdatedAtCell (the text difference is expected and unavoidable).
- Stop reading `window.innerWidth` in a render-time useState initializer in GlobalNav; initialize to 0 and set the real width in the existing effect.

Note: DateCell renders date-only calendar values (action start/end dates), which have no time zone. Those are formatted directly with `dayjs(date).format('L')`, which is already deterministic across environments; applying `.tz()` there would re-parse in the local zone and could shift the day.

Fixes WATCH-UI-3Q6

### DIFF
--- a/src/app/root/[domain]/[lang]/[plan]/layout.tsx
+++ b/src/app/root/[domain]/[lang]/[plan]/layout.tsx
@@ -22,6 +22,7 @@ import GlobalIndicatorModalWrapper from '@/components/indicators/GlobalIndicator
 import PathsProvider from '@/components/providers/PathsProvider';
 import PlanProvider from '@/components/providers/PlanProvider';
 import ThemeProvider from '@/components/providers/ThemeProvider';
+import TimeZoneProvider from '@/components/providers/TimeZoneProvider';
 import { SELECTED_WORKFLOW_COOKIE_KEY } from '@/constants/workflow';
 import { PrintProvider } from '@/context/print';
 import { WorkflowProvider } from '@/context/workflow-selector';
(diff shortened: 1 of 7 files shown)


woodwoerk (approved):
@bbliem The code changes aren't huge but this introduces potential for different timezones for the same user which is what I'm concerned about. If I understand correctly a user in Helsinki viewing a plan in California will see some dates/ times in EET and others in PT without knowing they're presented in different timezones.

However, I thought this was changing the timezone of all dates/times when reading the description yesterday so it's not as big a change as I thought. Let's go with it. 

For future hydration errors we can also consider just rendering text client-side only. If it's not vital for search engines and it can be rendered client-side cleanly without page jumps then that's often simpler.

Anyway, nice to get some of these hydration warnings fixed! Thanks!

PR #791, opened 2026-07-22
Make action cards centre aligned when smaller than a row's width

woodwoerk:
## Description

Small visual update to centre align action cards when there are too few to fit in a single row. This caused the cards to sit off to the left which looked odd against the centre aligned title. I used auto margins rather than flex centre aligning so that when on multiple rows, the cards are still left aligned, which is better for readability of these smaller action cards and is consistent with the action page.

### Before
image

### After
Screenshot 2026-07-22 at 14 51 23

Screenshot 2026-07-22 at 14 51 35

## Related issue

[Discussion in Slack](https://kausaltech.slack.com/archives/C029WNCMF0T/p1784547159807709)

---

### Type of Change

- [x] Set the PR's label to match the nature of this change

### Testing

- [-] **Built E2E tests** (if applicable. E2E tests added/updated)

- [-] **New strings are translatable** (all user-facing text uses i18n)

### Dependencies

### DIFF
--- a/src/components/actions/ActionCardList.tsx
+++ b/src/components/actions/ActionCardList.tsx
@@ -44,6 +44,21 @@ const ActionGroup = styled(Row)`
 const ActionGroupList = styled(Row)`
   padding: 0;
 `;
+
+// Centre the cards when there are fewer cards that fit into a row
+// to avoid a center aligned title and a few cards dangling off to the left.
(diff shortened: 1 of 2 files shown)


terotik (commented):
😽👌

terotik (approved):

